Issue #35, opened 2019-06-25
Remove Images and then Reinstall breaks Start command


tetchel:
@sujeilyfonseca got the installer into a broken state by installing, removing images and installing again, on Windows 10.

Now Codewind cannot start:

```
PS C:\Users\Administrator\.vscode\extensions\ibm.codewind-0.2.0\bin\installer\windows> .\codewind-installer.exe start
==> created file installer-docker-compose.yaml
==> environment structure written to installer-docker-compose.yaml
System architecture is:  amd64
Host operating system is:  windows
Please wait whilst containers initialize...
Recreating 6cb72082e843_codewind-performance ...
[1Beating 6cb72082e843_codewind-performance ... [31merror[0m
ERROR: for 6cb72082e843_codewind-performance  b'no such image: sha256:2b8b81d9edfd1216a7dbaaecd9b621cfa22f4e76d7904a44719492bfcec1391e: No such image: sha256:2b8b81d9edfd1216a7dbaaecd9b621cfa22f4e76d7904a44719492bfcec1391e'

ERROR: for codewind-performance  b'no such image: sha256:2b8b81d9edfd1216a7dbaaecd9b621cfa22f4e76d7904a44719492bfcec1391e: No such image: sha256:2b8b81d9edfd1216a7dbaaecd9b621cfa22f4e76d7904a44719492bfcec1391e'
The image for the service you're trying to recreate has been removed. If you continue, volume data could be lost. Consider backing up your data before continuing.

Continue with the new image? [yN]no such image: sha256:2b8b81d9edfd1216a7dbaaecd9b621cfa22f4e76d7904a44719492bfcec1391e: No such image: sha256:2b8b81d9edfd1216a7dbaaecd9b621cfa22f4e76d7904a44719492bfcec1391e
==> finished deleting file installer-docker-compose.yaml
Waiting for Codewind to start...
Waiting for Codewind to start...
Waiting for Codewind to start...
Waiting for Codewind to start...
Waiting for Codewind to start...
Waiting for Codewind to start...
Waiting for Codewind to start...
Waiting for Codewind to start...
Waiting for Codewind to start...
Waiting for Codewind to start...
Waiting for Codewind to start...
Waiting for Codewind to start...
Waiting for Codewind to start...
Waiting for Codewind to start...
Waiting for Codewind to start...
Codewind containers are taking a while to start. Please check the container logs and/or restart Codewind
[...]

PR #39, opened 2019-06-26
Exit health endpoint ping

liamchampton:
Problem: issue #35 
If docker-compose takes too long or fails to start up pfe, the pinging of the environment endpoint will exit without an error code at the end of the loop causing the plugin to take in a false positive.

Solution: (short term)
- At the end of the loop it will exit with a status code 1 to indicate an error for the plugins
- Increased loop timeout to help prevent hitting this
- A better solution of a ready/health endpoint API from codewind is needed


tetchel:
💯

tetchel:
fixes https://github.com/eclipse/codewind-installer/issues/35

PR #49, opened 2019-07-08
Removing artifactory image info

sishida:
Signed-off-by: Sarah Ishida <sishida@us.ibm.com>


tetchel:
this seems kind of disingenuous given that we are still using artifactory, and the installer still has a command to install from artifactory

sishida:
Closing this PR. We can work on more edits after support for artifactory is removed.

Issue #54, opened 2019-07-25
Add install flag to return unformatted terminal output


liamchampton:
**Subject of the issue**
The installer currently formats the json output of a docker pull so it is displayed nicely to the terminal. This is good if a user uses the installer manually but this is not good for the editor plugins as it restricts the progress display options.

**Currently**
```
latest: Pulling from ibmcom/codewind-pfe-amd64
8ba884070f61: Pull complete 
d29709d3e1de: Pull complete 
c01741a365bf: Pull complete 
01fc0d77508e: Pull complete 
...
```

**Flag needed to display**
```
c01741a365bf"}
{"status":"Downloading","progressDetail":{"current":531126,"total":75403831},"progress":"[\u003e                                                  ]  531.1kB/75.4MB","id":"8ba884070f61"}
{"status":"Downloading","progressDetail":{"current":801477,"total":6473076},"progress":"[======\u003e                                            ]  801.5kB/6.473MB","id":"d29709d3e1de"}
{"status":"Downloading","progressDetail":{"current":785093,"total":6103185},"progress":"[======\u003e                                            ]  785.1kB/6.103MB","id":"c01741a365bf"}
{"status":"Downloading","progressDetail":{"current":1592005,"total":6473076},"progress":"[============\u003e  
...
```

tetchel:
fixed #55

PR #55, opened 2019-07-25
add --json flag to the install command

liamchampton:
**Problem: #54**
The installer needs a flag option to return unformatted json. This will allow the editor plugins to have more flexibility when displaying progress during codewind installation.

**Solution:**
Added `--json/-j` flag on the install command e.g `./codewind-installer install --json` to get the raw json output

**Tested**
1. Manually tested all commands
2. Bats test output:
```
 ✓ invoke install command - using -t & -j flag
 ✓ invoke start command - Start dockerhub images'
 ✓ invoke stop-all command - Stop dockerhub images'
 ✓ invoke remove command - remove dockerhub images

4 tests, 0 failures
```

### DIFF
--- a/actions/commands.go
+++ b/actions/commands.go
@@ -53,6 +53,10 @@ func Commands() {
 					Value: "latest",
 					Usage: "dockerhub image tag",
 				},
+				cli.StringFlag{
+					Name:  "output, o",
+					Usage: "specify json terminal output",
+				},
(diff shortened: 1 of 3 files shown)


tetchel:
if `json` is the only possible value, is there a need to pass that value? can we just make it a flag (like a boolean) ?

PR #59, opened 2019-08-07
Clone repository using archive link

liamchampton:
**Enhancement needed**
Be able to clone a repository from github into a directory without using git

**Solution**
1. Get the archive link by querying the GitHub api
2. Download the files into `.zip` format into the tmp/temp dir of host machine by using the archive link
3. Unzip files into specified directory
4. Delete the `.zip` file stored in the tmp/temp dir

UPDATE
5. Updated packages to latest versions

**Commands**
1. `--branch/-b` **optional**
2. `--destination/-d` **required**
3. `--owner` **required**
4. `--repo` **required**

**Tested**
1. bats automated test. Output:
```
✓ invoke install command - using -t
 ✓ invoke start command - Start dockerhub images'
 ✓ invoke stop-all command - Stop dockerhub images'
 ✓ invoke remove command - remove dockerhub images
 ✓ invoke clone command - clone  microclimate-dev2ops/microclimateGoTemplate into ./clonetest

5 tests, 0 failures
```
2. Manually tested Mac/Windows environments

### DIFF
--- a/actions/clone-template.go
+++ b/actions/clone-template.go
@@ -142,13 +142,10 @@ func UnZip(zipFileName, destination string) {
 
 //MoveFiles to directory specified in command
 func MoveFiles(source, destination string) {
-	src := source
-	dest := destination
+	fmt.Println("==> moving files from ", source)
+	fmt.Println("==> moving files too ", destination)
 
-	fmt.Println("==> moving files from ", src)
-	fmt.Println("==> moving files too ", dest)
-
-	err := os.Rename(src, dest)
+	err := os.Rename(source, destination)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/actions/clone-template.go
+++ b/actions/clone-template.go
@@ -12,28 +12,142 @@
 package actions
 
 import (
-	"bytes"
+	"archive/zip"
+	"context"
 	"fmt"
+	"io"
 	"log"
-	"os/exec"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
 
+	"github.com/eclipse/codewind-installer/errors"
+	"github.com/eclipse/codewind-installer/utils"
+	"github.com/google/go-github/github"
 	"github.com/urfave/cli"
 )
 
-//CloneTemplate from github
+// CloneTemplate from github
 func CloneTemplate(c *cli.Context) {
 	//TODO Use go-git to do this in future
-	url := c.String("template")
-	if url == "" {
-		log.Fatal("No repository URL provided")
-	}
-	cmd := exec.Command("git", "clone", url)
-	output := new(bytes.Buffer)
-	cmd.Stdout = output
-	cmd.Stderr = output
-	if err := cmd.Start(); err != nil {
+	destination := c.String("destination")
+
+	zipURL := GetZipURL(c)
+	tempName := "/tmp/test_" + time.Now().String()
+	zipFileName := tempName + ".zip"
+
+	// download files in zip format
+	if err := DownloadFile(zipFileName, zipURL); err != nil {
 		log.Fatal(err)
 	}
-	cmd.Wait()
-	fmt.Printf(output.String())
+
+	// unzip into /tmp dir
+	extracedFilePath := UnZip(zipFileName, "/tmp")
+
+	// get top level dir from unzipped file and move files to new destination
+	if strings.HasPrefix(extracedFilePath, "/tmp/") {
+		extracedFilePath = strings.TrimPrefix(extracedFilePath, "/tmp/")
+	}
+	filePath := strings.Split(extracedFilePath, "/")
+	source := "/tmp/" + filePath[0]
+	MoveFiles(source, destination)
+
+	//delete zip file
+	utils.DeleteTempFile(zipFileName)
+
+}
+
+//GetZipURL from github api /repos/:owner/:repo/:archive_format/:ref
+func GetZipURL(c *cli.Context) string {
+	branch := c.String("branch")
+	owner := c.String("owner")
+	repo := c.String("repo")
+
+	client := github.NewClient(nil)
+
+	opt := &github.RepositoryContentGetOptions{Ref: branch}
+
+	URL, _, err := client.Repositories.GetArchiveLink(context.Background(), owner, repo, "zipball", opt)
+	if err != nil {
+		log.Fatal(err)
+	}
+	url := URL.String()
+	fmt.Println(URL)
+	return url
+}
+
+//DownloadFile - the git zip of master into current dir
+func DownloadFile(zipFileName, url string) error {
+
+	// Get the data
+	resp, err := http.Get(url)
+	errors.CheckErr(err, 400, "")
+	defer resp.Body.Close()
+
+	// Create the file
+	out, err := os.Create(zipFileName)
+	errors.CheckErr(err, 401, "")
+	defer out.Close()
+
+	// Write body to file
+	_, err = io.Copy(out, resp.Body)
+	fmt.Println(zipFileName)
+
+	return err
+}
+
+//UnZip downloaded file
+func UnZip(zipFileName, tempDestination string) string {
+	zipReader, _ := zip.OpenReader(zipFileName)
+
+	var extractedFilePath = ""
+	for _, file := range zipReader.Reader.File {
+
+		zippedFile, err := file.Open()
+		errors.CheckErr(err, 402, "")
+		defer zippedFile.Close()
+
+		targetDir := tempDestination
+		extractedFilePath = filepath.Join(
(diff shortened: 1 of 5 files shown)


utils/filesystem.go
```diff
@@ -104,3 +111,81 @@ func PingHealth(healthEndpoint string) bool {
 	}
 	return started
 }
+
+//GetZipURL from github api /repos/:owner/:repo/:archive_format/:ref
```
jcockbain96: Could this comment be a bit more descriptive, if you're going to have it?
I also personally don't see the purpose of writing the function name in the comment, as it's obvious what it's for, is there a GO reason for that?

utils/filesystem.go
```diff
@@ -104,3 +111,81 @@ func PingHealth(healthEndpoint string) bool {
...
+	URL, _, err := client.Repositories.GetArchiveLink(context.Background(), owner, repo, "zipball", opt)
+	if err != nil {
+		log.Fatal(err)
+	}
+	url := URL.String()
+	fmt.Println(URL)
+	return url
+}
+
+//DownloadFile - the git zip of master into current dir
```
jcockbain96: Similar to the above comment about this function comment.

actions/clone-template.go
```diff
@@ -0,0 +1,54 @@
...
+	} else {
+		tempPath = "/tmp/"
+	}
+
+	destination := c.String("destination")
+	branch := c.String("branch")
+	zipURL := utils.GetZipURL(c)
+	time := time.Now().Format(time.RFC3339)
+	time = strings.Replace(time, ":", "-", -1) // ":" is illegal char in windows
+	tempName := tempPath + branch + "_" + time
```
jcockbain96: This is a path, right? If this is the case, it would be clearer to me if the variable name referred to that, especially as that then bubbles down into downloadFile

utils/filesystem.go
```diff
@@ -104,3 +111,81 @@ func PingHealth(healthEndpoint string) bool {
...
+func DownloadFile(zipFileName, url string) error {
+
+	// Get the data
+	resp, err := http.Get(url)
+	errors.CheckErr(err, 400, "")
+	defer resp.Body.Close()
+
+	// Create the file
+	out, err := os.Create(zipFileName)
```
jcockbain96: Is `out` a variable name you're picking? Could that be something clearer?

actions/clone-template.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/eclipse/codewind-installer/utils"
+	"github.com/urfave/cli"
+)
+
+// CloneTemplate from github
+func CloneTemplate(c *cli.Context) {
+	var tempPath = ""
+
+	const GOOS string = runtime.GOOS
+	if GOOS == "windows" {
```
jcockbain96: Could the filepath package you've used below be used for this purpose? Just wondering, as path.join in node I think takes OS into account

utils/filesystem.go
```diff
@@ -104,3 +111,81 @@ func PingHealth(healthEndpoint string) bool {
...
+	client := github.NewClient(nil)
+
+	opt := &github.RepositoryContentGetOptions{Ref: branch}
+
+	URL, _, err := client.Repositories.GetArchiveLink(context.Background(), owner, repo, "zipball", opt)
+	if err != nil {
+		log.Fatal(err)
+	}
+	url := URL.String()
+	fmt.Println(URL)
```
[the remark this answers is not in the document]
jcockbain96: You've done it below as well, so I'm guessing there's a reason for it

utils/filesystem.go
```diff
@@ -104,3 +111,81 @@ func PingHealth(healthEndpoint string) bool {
...
+
+	// Write body to file
+	_, err = io.Copy(out, resp.Body)
+	fmt.Println(zipFileName)
+
+	return err
+}
+
+//UnZip downloaded file
+func UnZip(zipFileName, destination string) {
```
jcockbain96: As you only refer to it once, could destination be a more descriptive variable name? Just to make it easier to follow

errors/error.go
```diff
@@ -55,6 +55,16 @@ func CheckErr(err error, code int, optMsg string) {
 			log.Print("DELETE_FILE_ERROR", "[", code, "]: ", err, ". ", optMsg)
 		case 300:
 			log.Fatal("APPLICATION_ERROR", "[", code, "]: ", err, ". ", optMsg)
+		case 400:
+			log.Fatal("REPOSITORY_DOWNLOAD_ERROR", "[", code, "]: ", err, ". ", optMsg)
+		case 401:
+			log.Fatal("CREATE_ZIP_FILE_ERROR", "[", code, "]: ", err, ". ", optMsg)
+		case 402:
+			log.Fatal("ZIP_FILE_READ_ERROR", "[", code, "]: ", err, ". ", optMsg)
```
jcockbain96: Might be nice to be consistent here and have `READ_ZIP_FILE_ERROR` to go with `CREATE_ZIP_FILE_ERROR`?

utils/filesystem.go
```diff
@@ -104,3 +111,81 @@ func PingHealth(healthEndpoint string) bool {
...
+	zipReader, _ := zip.OpenReader(zipFileName)
+
```
jcockbain96: Also, would you say this function is doing more than unzipping? Is it copying a zipped file to the file system  destination? Maybe the function name should reflect that?

actions/clone-template.go
liamchampton: Not sure filepath can be used. The runtime is the safest way to determine the operating system in this case.

actions/clone-template.go
```diff
@@ -0,0 +1,54 @@
...
+		tempPath = os.Getenv("TEMP") + "\\"
+	} else {
+		tempPath = "/tmp/"
+	}
+
+	destination := c.String("destination")
```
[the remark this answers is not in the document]
liamchampton: Sure. Will break the variable declarations away from function calls

errors/error.go
liamchampton: Good spot.

utils/filesystem.go
liamchampton: Go linter. These are global functions so it wants/needs comments. I would prefer to keep these comments brief with just the key elements if possible - I dont think it really needs much more than this. Would you have a suggestion?

utils/filesystem.go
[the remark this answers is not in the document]
liamchampton: > Will this just print the URL out without context?

- Can add a word or two before this to give context for sure. Don't want to bloat the response anymore than necessary this balance is hard with a cli! 


> You've done it below as well, so I'm guessing there's a reason for it

- Done what below? I return the url string, not print it out

utils/filesystem.go
liamchampton: Can probably make this a bit clearer

utils/filesystem.go
liamchampton: sure, will change to be clearer

utils/filesystem.go
liamchampton: No, this is just unzipping the contents of the zip file into the destination specified by the user. The range loop is iterating over each element within the zip file and putting it into the correct directory. As attempted before, unzipping into the temp directory provides more issues when it comes to moving the files on the file system. (see previous commits)

utils/filesystem.go
liamchampton: sure!

actions/clone-template.go
liamchampton: This is a file name -`tempPath` is a path

liamchampton:
Bats test passing after pr review changes and rebase
```
bats integration.bats
 ✓ invoke install command - using -t
 ✓ invoke start command - Start dockerhub images'
 ✓ invoke stop-all command - Stop dockerhub images'
 ✓ invoke remove command - remove dockerhub images
 ✓ invoke clone command - clone  microclimate-dev2ops/microclimateGoTemplate into ./clonetest

5 tests, 0 failures
```

utils/filesystem.go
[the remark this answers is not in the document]
jcockbain96: fmt.Println(...) before each return statement is what I was referring too, have seen you've added some context 👍

utils/filesystem.go
jcockbain96: `GetZipUrl from the github API, with format /repos/:owner/:repo/:archive_format/:ref` something like that? Comments are fine, I found this one not as clear as some of your others

utils/filesystem.go
jcockbain96: Yeah sorry, my comment was badly written. I'd still say this isn't an easy function to understand. Maybe something like `Unzip ToDestination(...)`. Not sure though

PR #70, opened 2019-08-13
add project validation on project command

jcockbain96:

**Problem** 

Project validation is currently performed by the initialize container within pfe, this includes determining the project language and buildType, and writing a .cw-settings file to the project. We are getting rid of this container as we move forwards towards a Hybrid architecture. This is because we won't have direct access to a users file system.

**Solution**

This moves the determining of language and buildType to this cli, with the writing of the settings to come. The same JSON is returned from the command ./installer <project-path> as was by the /validate API on PFE.

**Tested** 

Manual so far, adding unit tests now.

### DIFF
--- a/actions/commands.go
+++ b/actions/commands.go
@@ -56,7 +56,7 @@ func Commands() {
 		{
 			Name:    "install",
 			Aliases: []string{"in"},
-			Usage:   "Pull pfe, performance & intialize images from dockerhub",
+			Usage:   "Pull pfe, performance & initialize images from dockerhub",
 			Flags: []cli.Flag{
 				cli.StringFlag{
 					Name:  "tag, t",
--- a/actions/commands.go
+++ b/actions/commands.go
@@ -44,11 +44,10 @@ func Commands() {
 				},
 			},
 			Action: func(c *cli.Context) error {
-				if c.NumFlags() == 0 {
-					ValidateProject(c)
-				} else {
+				if c.NumFlags() != 0 {
 					DownloadTemplate(c)
 				}
+				ValidateProject(c)
 				return nil
 			},
 		},
--- a/actions/commands.go
+++ b/actions/commands.go
@@ -45,7 +45,7 @@ func Commands() {
 			},
 			Action: func(c *cli.Context) error {
 				if c.NumFlags() == 0 {
-					// TODO: add ValidateProject() func
+					ValidateProject(c)
 				} else {
 					DownloadTemplate(c)
 				}
--- a/actions/project.go
+++ b/actions/project.go
@@ -14,7 +14,6 @@ package actions
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"log"
 	"os"
 	"path"
@@ -39,17 +38,6 @@ type (
 		Status string      `json:"status"`
 		Path   string      `json:"path"`
 		Result ProjectType `json:"result"`
-	}
-
-	// CWSettings represents the .cw-settings file which is written to a project
-	CWSettings struct {
-		ContextRoot       string   `json:"contextRoot"`
-		InternalPort      string   `json:"internalPort"`
-		HealthCheck       string   `json:"healthCheck"`
-		InternalDebugPort *string  `json:"internalDebugPort"`
-		IgnoredPaths      []string `json:"ignoredPaths"`
-		MavenProfiles     []string `json:"mavenProfiles,omitempty"`
-		MavenProperties   []string `json:"mavenProperties,omitempty"`
 	}
 )
 
@@ -100,12 +88,13 @@ func DownloadTemplate(c *cli.Context) {
 	utils.DeleteTempFile(zipFileName)
 }
 
-// ValidateProject returns the language and buildType for a project at given filesystem path
+// ValidateProject returns the language and buildType for a project at given filesystem path,
+// and writes a default .cw-settings file to that project
 func ValidateProject(c *cli.Context) {
 	projectPath := c.Args().Get(0)
-	checkProjectPath(projectPath)
+	utils.CheckProjectPath(projectPath)
 
-	language, buildType := determineProjectInfo(projectPath)
+	language, buildType := utils.DetermineProjectInfo(projectPath)
 	response := ValidationResponse{
 		Status: "success",
 		Result: ProjectType{language, buildType},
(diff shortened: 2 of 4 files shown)


actions/project.go
```diff
@@ -69,7 +86,56 @@ func DownloadTemplate(c *cli.Context) {
...
+	language, buildType := determineProjectInfo(projectPath)
+
+	resp := ValidationResponse{
+		Status: "success",
+		Result: ProjectType{language, buildType},
+		Path:   projectPath,
+	}
+
+	projectInfo, err := json.Marshal(resp)
+	if err != nil {
```
liamchampton: Can this use the error checking function set up?

actions/project.go
```diff
@@ -69,7 +86,56 @@ func DownloadTemplate(c *cli.Context) {
...
+		log.Fatal(err)
+		return
+	}
+	fmt.Println(string(projectInfo))
+}
+
+func determineProjectInfo(projectPath string) (string, string) {
+	language := "unknown"
+	buildType := "docker"
+	if _, err := os.Stat(path.Join(projectPath, "pom.xml")); err == nil {
```
liamchampton: Could this possibly be a switch statement? - might look neater and easier to maintain

actions/project.go
```diff
@@ -69,7 +86,56 @@ func DownloadTemplate(c *cli.Context) {
...
+
+	projectInfo, err := json.Marshal(resp)
+	if err != nil {
+		log.Fatal(err)
+		return
+	}
+	fmt.Println(string(projectInfo))
+}
+
+func determineProjectInfo(projectPath string) (string, string) {
```
liamchampton: Too keep this in line with other functions and the possibility that this could be called from outside of this local scope in future, could this be global? i.e `DetermineProjectInfo`?

jcockbain96: Yeah sure. As I would with node, I only exported what had to be exported. Wouldn't you not make it global only if it became necessary to?

actions/project.go
```diff
@@ -69,7 +97,99 @@ func DownloadTemplate(c *cli.Context) {
 	utils.DeleteTempFile(zipFileName)
 }
 
-//ValidateProject type
-func ValidateProject() {
-	//code here
+// ValidateProject returns the language and buildType for a project at given filesystem path
+func ValidateProject(c *cli.Context) {
+	projectPath := c.Args().Get(0)
```
hhellyer: DownloadTemplate() has a check that the directory parameter is set. This function should have one too!
See: https://github.com/eclipse/codewind-installer/pull/70/files#diff-7cf8a504cf7756979e16701b4b2789b7R57

actions/commands.go
```diff
@@ -45,7 +45,7 @@ func Commands() {
 			},
 			Action: func(c *cli.Context) error {
 				if c.NumFlags() == 0 {
-					// TODO: add ValidateProject() func
+					ValidateProject(c)
 				} else {
 					DownloadTemplate(c)
 				}
```
hhellyer: Don't we want to always call `ValidateProject()` even if we call `DownloadTempalate()` first?

actions/project.go
hhellyer: Keeping it local until we need to expose it further would probably be best.

actions/project.go
jcockbain96: What would you switch on?

utils/project.go
```diff
@@ -0,0 +1,129 @@
...
+		buildType = "swift"
+	}
+	return language, buildType
+}
+
+// CheckProjectPath will stop the process and return an error if path does not
+// exist or is invalid
+func CheckProjectPath(projectPath string) {
+	if projectPath == "" {
+		log.Fatal("Project path has not been set")
```
rwalle61: comment: this function does 2 things - it validates the project path, and controls logic flow based on the result (i.e. if the path is invalid, this will call `log.Fatal`, which calls `os.exit`). If we ever want to `CheckProjectPath` without risking an `os.exit`, then we should separate the two. 

(This is a general point about our error handling in this CLI - a lot of functions throw errors rather than returning them, which means that if in future want to handle errors we will have to adapt our code. See https://stackoverflow.com/questions/33885235/should-a-go-package-ever-use-log-fatal-and-when )

utils/project.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	if _, err := os.Stat(projectPath); os.IsNotExist(err) {
+		log.Fatal("Project not found at given path")
+	}
+}
+
+func determineJavaBuildType(projectPath string) string {
+	pathToPomXML := path.Join(projectPath, "pom.xml")
+	pomXMLContents, _err := ioutil.ReadFile(pathToPomXML)
```
rwalle61: `_err` -> `err`?

utils/project.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// WriteNewCwSettings writes a default .cw-settings file to the given path,
+// dependant on the build type of the project
+func WriteNewCwSettings(pathToCwSettings string, BuildType string,) {
+	defaultCwSettings := getDefaultCwSettings()
+	cwSettings := addNonDefaultFieldsToCwSettings(defaultCwSettings, BuildType)
+	settings, err := json.MarshalIndent(cwSettings, "", "")
+	errors.CheckErr(err, 203, "")
+	err = ioutil.WriteFile(pathToCwSettings, settings, 0644)
```
rwalle61: could comment what permissions 0644 grants

utils/project.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// RenameLegacySettings renames a .mc-settings file to .cw-settings
+func RenameLegacySettings(pathToLegacySettings string, pathToCwSettings string) {
+	err := os.Rename(pathToLegacySettings, pathToCwSettings)
+	errors.CheckErr(err, 205, "")
+}
+
+func getDefaultCwSettings() CWSettings {
```
rwalle61: could this be a const instead?

utils/project.go
```diff
@@ -0,0 +1,129 @@
...
+		cwSettings.InternalDebugPort = &defaultValue
+	}
+	if stringInSlice(ProjectType, projectTypesWithMavenSettings) {
+		cwSettings.MavenProfiles = []string{""}
+		cwSettings.MavenProperties = []string{""}
+	}
+	return cwSettings
+}
+
+func stringInSlice(a string, list []string) bool {
```
rwalle61: nit: given this is called `stringInSlice`, maybe `list` should be renamed to `slice`?

actions/project.go
```diff
@@ -69,7 +88,32 @@ func DownloadTemplate(c *cli.Context) {
...
+	response := ValidationResponse{
+		Status: "success",
+		Result: ProjectType{language, buildType},
+		Path:   projectPath,
+	}
+	projectInfo, err := json.Marshal(response)
+
+	errors.CheckErr(err, 203, "")
+	writeCwSettingsIfNotInProject(projectPath, buildType)
+	fmt.Println(string(projectInfo))
```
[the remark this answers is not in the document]
jcockbain96: A stringified json of the from 
```
{ 
	status: "success",
	path: "/Users/james/python-project,
	result: {
		language: "java",
		buildType: "python",
	},
}
```

utils/project.go
```diff
@@ -0,0 +1,129 @@
...
+	InternalDebugPort *string  `json:"internalDebugPort"`
+	IgnoredPaths      []string `json:"ignoredPaths"`
+	MavenProfiles     []string `json:"mavenProfiles,omitempty"`
+	MavenProperties   []string `json:"mavenProperties,omitempty"`
+}
+
+// DetermineProjectInfo returns the language and build-type of a project
+func DetermineProjectInfo(projectPath string) (string, string) {
+	language, buildType := "unknown", "docker"
+	if _, err := os.Stat(path.Join(projectPath, "pom.xml")); err == nil {
```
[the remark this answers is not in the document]
jcockbain96: Nice, yeah that would be useful in a few places

utils/project.go
jcockbain96: A global const?

Issue #69, opened 2019-08-13
Restrict PFE port range to avoid clashes with application port usage


mattcolegate:
Auto-assigning ports for the PFE container has revealed a problem in the assigning of the port to be used.
As PFE specifies the hostname interface to be used as 127.0.0.1 (loopback or localhost) we just bind to the port on that interface (say, 32770).
When an application is created, this binds to all interfaces on the first available port, using the IP shorthand 0.0.0.0, This will bind to the public interfaces of any attached hardware, but is unable to bind on localhost because PFE got there first. The port was still marked as available because PFE asked for it just on localhost, not on 0.0.0.0.
We can't use 0.0.0.0 for PFE because that would then expose it on all interfaces, including public ones, creating a security omnishambles. Therefore in order to fix this without losing the dynamic port assignment, we will have PFE try to connect on a range of ports - 10000 to 11000 - that avoids the ranges assigned by Docker (32700+) whilst retaining the ability to skip over any ports that might be in use by other processes.

liamchampton:
@mattcolegate this issue can be closed now

PR #71, opened 2019-08-13
Check for available TCP port in range

liamchampton:
**Problem #69**
There is an issue where docker will select and assign a random port, or at least attempt to. If the port is restricted or unreachable then it will fail to do so.

**Solution**
Write a function to check for a free port within the range specified `10000 -> 11000`. It will return a bool & port number. This will then allow docker to use it knowing it is available.

**Tested**
Manually - output:
```
2019/08/13 14:47:46 Port 10000 Available
```

**Extra**
Removed some white space elsewhere in the file

### DIFF
--- a/utils/docker.go
+++ b/utils/docker.go
@@ -328,7 +328,7 @@ func IsTCPPortAvailable() (bool, string) {
 			log.Println(status)
 		} else {
 			status = "Port " + strconv.Itoa(port) + " Available"
-			log.Println(status)
+			fmt.Println(status)
 			conn.Close()
 			return true, strconv.Itoa(port)
 		}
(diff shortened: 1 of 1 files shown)


utils/docker.go
```diff
@@ -310,3 +312,26 @@ func GetPFEPort() string {
 	}
 	return ""
 }
+
+// IsTCPPortAvailable checks to see if a specified port is available
```
markcor11: should the comment indicate that the function will pick the next available port ?

liamchampton: You are correct, yes. It doesn't only find a port, it also returns it.

PR #80, opened 2019-08-19
report status of codewind

liamchampton:
**Problem https://github.com/eclipse/codewind/issues/176**
The end user does not currently have an easy way to tell which version of Codewind images are installed or started. The status command will tell the end user whether it is installed or not/what port it is running on but not the version of images.

**Solution**
Modify the json output of the status command by adding `installed-versions` and `started` arrays. This will notify the end user which versions of codewind images are installed on the host machine and which one is currently running. As per the issue, "apparently there may be cases in the future where we have multiple running versions, so started could be an array too." -`started` is an array to accommodate for this in the future. 

Installed but not running - `./installer status -j` => `{"status":"stopped","installed-versions":["latest","0.2"]}`
Installed and running - `./installer status -j` => `{"status":"started","url":"http://127.0.0.1:10000","installed-versions":["latest","0.2"],"started":["latest"]}`

**Tested**
- Unit test written to test removal of duplicate elements in an array. Output:
```
Running tool: /usr/local/bin/go test -timeout 30s github.com/eclipse/codewind-installer/utils -run ^(TestRemoveArrayDuplicate)$

ok  	github.com/eclipse/codewind-installer/utils	0.048s
Success: Tests passed.
```

- Bats testing all commands. Output:
```
➜  codewind-installer git:(status-tag) ✗ bats integration.bats 
 ✓ invoke install command - default to latest tag
 ✓ invoke status -j command - output = '{status:stopped,installed-versions:[latest]}'
 ✓ invoke start command - Start dockerhub images (latest)
 ✓ invoke stop-all command - Stop dockerhub images (latest)
 ✓ invoke remove command - remove all dockerhub images
 ✓ invoke project command - using -r download microclimate-dev2ops/microclimateGoTemplate into ./downloadtest
 ✓ cleanup

7 tests, 0 failures
```

### DIFF
--- a/utils/docker.go
+++ b/utils/docker.go
@@ -325,8 +325,8 @@ func GetPFEHostAndPort() (string, string) {
 	return "", ""
 }
 
-// GetImageTag of Codewind images
-func GetImageTag() []string {
+// GetImageTags of Codewind images
+func GetImageTags() []string {
 	imageArr := [3]string{}
 	imageArr[0] = "eclipse/codewind-pfe"
 	imageArr[1] = "eclipse/codewind-performance"
@@ -352,7 +352,7 @@ func GetImageTag() []string {
 		}
 	}
 
-	tagArr = RemoveArrayDuplicate(tagArr)
+	tagArr = RemoveDuplicateEntries(tagArr)
 	return tagArr
 }
 
@@ -373,8 +373,8 @@ func IsTCPPortAvailable(minTCPPort int, maxTCPPort int) (bool, string) {
 	return false, ""
 }
 
-// GetContainerTag of the Codewind version(s) currently running
-func GetContainerTag() []string {
+// GetContainerTags of the Codewind version(s) currently running
+func GetContainerTags() []string {
 	containerArr := [2]string{}
 	containerArr[0] = "codewind-pfe"
 	containerArr[1] = "codewind-performance"
@@ -391,6 +391,6 @@ func GetContainerTag() []string {
 		}
 	}
 
-	tagArr = RemoveArrayDuplicate(tagArr)
+	tagArr = RemoveDuplicateEntries(tagArr)
 	return tagArr
 }
--- a/utils/docker.go
+++ b/utils/docker.go
@@ -358,25 +358,8 @@ func GetImageTag() []string {
 		}
 	}
 
-	//fmt.Println("FULL tag array = ", tagArr)
-
-	// Use map to record duplicates if found
-	encountered := map[string]bool{}
-	result := []string{}
-
-	for element := range tagArr {
-		if encountered[tagArr[element]] == true {
-			// Don't add duplicate
-		} else {
-			// Record said element as an encountered
-			encountered[tagArr[element]] = true
-			// Append to the new result slice
-			result = append(result, tagArr[element])
-		}
-	}
-
-	//fmt.Println("refactored result =  ", result)
-	return result
+	tagArr = RemoveArrayDuplicate(tagArr)
+	return tagArr
 }
 
 // IsTCPPortAvailable checks to find the next available port and returns it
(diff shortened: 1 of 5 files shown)


tetchel:
this look spectacular; please have a look at the jenkins build failure. 

```
10:34:30  Gopkg.lock is out of sync with imports and/or Gopkg.toml. Run `dep check` for details.
10:34:30  PROJECT                                  MISSING PACKAGES
10:34:30  github.ibm.com/codewind-installer/utils  [github.ibm.com/codewind-installer/utils]  
10:34:30  input-digest mismatch
```

utils/utils.go
```diff
@@ -0,0 +1,32 @@
...
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+
+package utils
+
+// RemoveArrayDuplicate elements
+func RemoveArrayDuplicate(tagArr []string) []string {
```
markcor11: Could this be renamed to RemoveDuplicateEntries to make clear what its doing ?

utils/utils.go
```diff
@@ -0,0 +1,32 @@
...
+	// Use map to record duplicates if found
+	encountered := map[string]bool{}
+	result := []string{}
+
+	for element := range tagArr {
+		if encountered[tagArr[element]] == true {
```
markcor11: Perhaps just test for false and remove the empty code block ?   Perhaps it might also be cleaner to use a map and store the elements as the keys which would make them unique, then write the map out as a new array.

utils/docker.go
```diff
@@ -322,7 +322,38 @@ func GetPFEHostAndPort() (string, string) {
 			}
 		}
 	}
-	return "",""
+	return "", ""
+}
+
+// GetImageTag of Codewind images
+func GetImageTag() []string {
```
rwalle61: This should be plural right?

```suggestion
func GetImageTags() []string {
```

utils/docker.go
```diff
@@ -341,3 +372,25 @@ func IsTCPPortAvailable(minTCPPort int, maxTCPPort int) (bool, string) {
 	}
 	return false, ""
 }
+
+// GetContainerTag of the Codewind version(s) currently running
+func GetContainerTag() []string {
```
rwalle61: This should also be pluralised?

utils/utils.go
rwalle61: and this works for slices too right, so `tagArr` -> `slice`?

utils/utils.go
```diff
@@ -0,0 +1,32 @@
...
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+
+package utils
+
+// RemoveArrayDuplicate elements
+func RemoveArrayDuplicate(tagArr []string) []string {
+	// Use map to record duplicates if found
+	encountered := map[string]bool{}
```
rwalle61:
```suggestion
	encounteredElements := map[string]bool{}
```

utils/utils.go
liamchampton: Reworked this logic to make better use of the map

utils/utils.go
liamchampton: Changed to `encounteredElement`

liamchampton:
**Update** - bats and unit test still passing

utils/utils_test.go
```diff
@@ -107,3 +107,24 @@ func TestDeleteTempFileFail(t *testing.T) {
...
+		log.Fatal("Test 1: Failed to delete duplicate array index")
+	}
+
+	dupArr = []string{"", "test", "test"}
+	result = RemoveDuplicateEntries(dupArr)
+	if len(result) != 1 {
+		log.Fatal("Test 2: Failed to delete duplicate array index")
+	}
+
+	dupArr = []string{"", "", ""}
```
rwalle61: Nit: should `RemoveDuplicateEntries(["", "", ""])` really return `[]`, rather than `[""]`?

liamchampton: It does return []. If the length of an array = 0 then there are no elements. The return value is [] and not [""] as this would then be 1.

utils/utils.go
markcor11: Looks better thanks!

PR #81, opened 2019-08-19
Add first templates command to list templates

rwalle61:
## Problem
To manage project templates, Codewind Vscode plugins want to call this Go CLI rather than Codewind's REST API directly. 

## Solution
Add the first `codewind-installer templates` command. (This will be followed by further PRs).

`codewind-installer templates list` calls Codewind's REST API `GET templates` to output:
image

Covered by a test.

### DIFF
--- a/actions/commands.go
+++ b/actions/commands.go
@@ -143,7 +143,7 @@ func Commands() {
 				{
 					Name:  "list",
 					Aliases: []string{"ls"},
-					Usage: "list available template repositories",
+					Usage: "list available templates",
 					Action: func(c *cli.Context) error {
 						ListTemplates()
 						return nil
--- a/actions/commands.go
+++ b/actions/commands.go
@@ -135,6 +135,22 @@ func Commands() {
 				return nil
 			},
 		},
+
+		{
+			Name:    "templates",
+			Usage:   "Manage project templates",
+			Subcommands: []cli.Command{
+				{
+					Name:  "list",
+					Aliases: []string{"ls"},
+					Usage: "list available template repositories",
+					Action: func(c *cli.Context) error {
+						ListTemplates()
+						return nil
+					},
+				},
+			},
+		},
 	}
 
 	// Start application
--- a/actions/templates.go
+++ b/actions/templates.go
@@ -20,7 +20,7 @@ import (
 	"github.com/eclipse/codewind-installer/config"
 )
 
-// Template represents a template.
+// Template represents a project template.
 type Template struct {
 	Label       string `json:"label"`
 	Description string `json:"description"`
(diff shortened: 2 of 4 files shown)


actions/commands.go
```diff
@@ -135,6 +135,22 @@ func Commands() {
 				return nil
 			},
 		},
+
+		{
+			Name:    "templates",
+			Usage:   "Manage project templates",
+			Subcommands: []cli.Command{
+				{
+					Name:  "list",
```
liamchampton: Are we going to have more sub commands for `templates`? If not, i would maybe have this as a flag?

actions/templates.go
```diff
@@ -0,0 +1,66 @@
...
+		return
+	}
+	PrettyPrintJSON(templates)
+}
+
+// GetTemplates extracts URLs from the file at the provided path,
+// then gets the template descriptions from those URLs,
+// then formats the descriptions into template objects.
+func GetTemplates() ([]Template, error) {
+	resp, err := http.Get("http://localhost:9090/api/v1/templates")
```
liamchampton: This wont be 9090 since it is now dynamically being assigned between 10000->11000. You will need to get the port number by calling `utils.GetPFEHostAndPort`

actions/commands.go
rwalle61: we will :)

actions/templates.go
```diff
@@ -0,0 +1,66 @@
...
+package actions
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+// Template represents a template.
```
jcockbain96:
```suggestion
// Template represents a project template.
```

actions/templates.go
```diff
@@ -0,0 +1,66 @@
...
+type Template struct {
+	Label       string `json:"label"`
+	Description string `json:"description"`
+	Language    string `json:"language"`
+	URL         string `json:"url"`
+	ProjectType string `json:"projectType"`
+}
+
+// ListTemplates lists all templates Codewind is aware of.
```
jcockbain96:
```suggestion
// ListTemplates lists all project templates Codewind is aware of.
```

actions/templates_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetTemplates(t *testing.T) {
+	tests := map[string]struct {
+		wantedType   []Template
+		wantedLength int
+		wantedErr    error
+	}{
+		"success case: no input": {
```
jcockbain96:
```suggestion
		"success case": {
```

actions/templates_test.go
```diff
@@ -0,0 +1,29 @@
...
+			wantedType:   []Template{},
+			wantedLength: 8,
```
jcockbain96: May become an issue when templates are added? Could we mock this list when we get time, which will be useful for all future template commands too?

actions/templates.go
rwalle61: Thanks, I've added a commit to do this

PR #86, opened 2019-08-21
Documentation update

liamchampton:
**Problem**
Now bats tests have gone into the repository there needs to be accompanying documentation on how to set it up on a machine. There is also no documentation of the actual commands available within in the installer. 

**Solution**
- add bats installation instructions
- add table of available commands followed by flags/subcommand information

**Extra**
- code formatting

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ $ brew upgrade dep
 
 Subcommands:</br>
 
-`list/ls` - Shows a list of commands or help for one command
+`list/ls` - List available templates
 
 ## help
 
--- a/README.md
+++ b/README.md
@@ -83,49 +83,58 @@ $ brew upgrade dep
 
 |Command         |Alias         |Usage                                                           |
 |----------------|--------------|----------------------------------------------------------------|
-|project         |-             |'Manage Codewind projects'                                      |
+|project         |`-`           |'Manage Codewind projects'                                      |
 |install         |`in`          |'Pull pfe, performance & initialize images from dockerhub'      |
-|start           |-             |'Start the Codewind containers'                                 |
-|status          |-             |'Print the installation status of Codewind'                     |
-|stop            |-             |'Stop the running Codewind containers'                          |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -67,10 +67,77 @@ $ brew upgrade dep
...
+
+## CLI Command Options
+
+## project
+
+`--url/-u <value>` - URL of project to download
+
+## install
+
+`--tag/t <value>` - Dockerhub image tag (default: "latest")</br>
```
markcor11: in most cases below you've got the abbreviation prefixed by a single dash (eg line 109).  This option seems to use t without a dash.

README.md
```diff
@@ -67,10 +67,77 @@ $ brew upgrade dep
...
+2. Clone the `codewind-installer` repository.
+3. Use the `cd` command to go to the top level project directory.
+4. Ensure your system environment is clean by having no Codewind images installed or containers running.
+5. To run the tests, enter the `bats integration.bats` command in the command line window and wait for the tests to finish.
+
+## CLI Commands
+
+|Command         |Alias         |Usage                                                           |
+|----------------|--------------|----------------------------------------------------------------|
+|project         |`-`           |'Manage Codewind projects'                                      |
```
markcor11: Not clear on how to use the dash alias here or in the lines below.  Are you saying that I can type a dash and that means project or means start or means stop (which seems confusing) or does it mean "Not applicable" in which case it may be better to just keep that cell in the table empty.

README.md
```diff
@@ -67,10 +67,77 @@ $ brew upgrade dep
...
+`--tag/-t <value>` - Dockerhub image tag.</br>
+**Note:** Failing to specify a `--tag`, will remove all Codewind images on the host machine.
+
+## templates
+
+>**Note:** No additional flags
+
+Subcommands:</br>
+
+`list/ls` - Shows a list of commands or help for one command
```
rwalle61: at least for `templates list`, `templates repos list`, and `templates styles list`:
`list` lists the relevant resources

PR #101, opened 2019-09-03
Fix up linter warnings

mattcolegate:
Getting rid of the linter warnings

Signed-off-by: Matthew Colegate <colegate@uk.ibm.com>


johnmcollier:
FWIW, probably a good idea to add a call to `golint` to the Jenkinsfile too, so that linter failures will cause builds to fail

PR #103, opened 2019-09-04
quickfix to remove Appsody containers on stop-all command

jcockbain96:
**Problem**
The stop-all command does stop Appsody project containers, as the code only checks for codewind prefixes on the image.
https://github.com/eclipse/codewind/issues/224

**Solution**
Add a check for `appsody` in the images (quickfix), this will also remove existing Appsody projects. A better solution will be required after ifix, one in which we can determine codewind Appsody projects only.

**Tests**
Manual

### DIFF
--- a/actions/remove.go
+++ b/actions/remove.go
@@ -22,16 +22,17 @@ import (
 //RemoveCommand to remove all codewind and project images
 func RemoveCommand(c *cli.Context) {
 	tag := c.String("tag")
-	imageArr := [4]string{}
+	imageArr := [5]string{}
 	imageArr[0] = "eclipse/codewind-pfe"
 	imageArr[1] = "eclipse/codewind-performance"
 	imageArr[2] = "eclipse/codewind-initialize"
 	imageArr[3] = "cw-"
+	imageArr[4] = "appsody"
 	networkName := "codewind"
 
 	if tag != "" {
 		for i := 0; i < len(imageArr); i++ {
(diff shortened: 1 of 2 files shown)


makandre:
I don't think we should remove appsody stack images.  Those are like base images that appsody applications use.  Deleting them would be like us deleting the base `node` image that we build our images from

When removing container, could we check for `appsody` image name *and* that the container name starts with `cw-`?  That makes it less likely we remove appsody containers started outside Codewind by mistake.

jcockbain96:
I agree we shouldn't - will add the container name check too.

actions/remove.go
```diff
@@ -22,16 +22,17 @@ import (
 //RemoveCommand to remove all codewind and project images
 func RemoveCommand(c *cli.Context) {
 	tag := c.String("tag")
-	imageArr := [4]string{}
+	imageArr := [5]string{}
 	imageArr[0] = "eclipse/codewind-pfe"
 	imageArr[1] = "eclipse/codewind-performance"
 	imageArr[2] = "eclipse/codewind-initialize"
 	imageArr[3] = "cw-"
+	imageArr[4] = "appsody"
```
makandre: Is this removing appsody images?

jcockbain96: It's not removing them, but wasn't supposed to be left in the list. https://github.com/eclipse/codewind-installer/pull/106 removes it.

PR #106, opened 2019-09-05
remove appsody from array of images to remove

jcockbain96:
The previous PR left in `appsody` in the array of prefixes to check when removing images. This doesn't actually remove images, but isn't needed and is confusing.


tobespc:
closing as an other change fixes this

PR #108, opened 2019-09-06
fix removing of images

tobespc:
Signed-off-by: Toby Corbin <corbint@uk.ibm.com>

fixed the logic for detecting and removing of images

https://github.com/eclipse/codewind/issues/389

### DIFF
--- a/actions/remove.go
+++ b/actions/remove.go
@@ -22,21 +22,13 @@ import (
 //RemoveCommand to remove all codewind and project images
 func RemoveCommand(c *cli.Context) {
 	tag := c.String("tag")
-	imageArr := [4]string{}
-	imageArr[0] = "eclipse/codewind-pfe"
-	imageArr[1] = "eclipse/codewind-performance"
-	imageArr[2] = "eclipse/codewind-initialize"
-	imageArr[3] = "cw-"
+	imageArr := []string{
+        "eclipse/codewind-pfe-amd64:" + tag,
+        "eclipse/codewind-performance-amd64:" + tag,
+        "eclipse/codewind-initialize-amd64:" + tag,
+        "cw-",
+    }
 	networkName := "codewind"
-
-	if tag != "" {
-		for i := 0; i < len(imageArr); i++ {
-			if i == 3 {
-				break
-			}
-			imageArr[i] = imageArr[i] + "-amd64:" + tag
-		}
-	}
 
 	images := utils.GetImageList()
 
--- a/actions/remove.go
+++ b/actions/remove.go
@@ -22,17 +22,16 @@ import (
 //RemoveCommand to remove all codewind and project images
 func RemoveCommand(c *cli.Context) {
 	tag := c.String("tag")
-	imageArr := [5]string{}
+	imageArr := [4]string{}
 	imageArr[0] = "eclipse/codewind-pfe"
 	imageArr[1] = "eclipse/codewind-performance"
 	imageArr[2] = "eclipse/codewind-initialize"
 	imageArr[3] = "cw-"
-	imageArr[4] = "appsody"
 	networkName := "codewind"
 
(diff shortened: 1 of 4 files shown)


actions/stop-all.go
```diff
@@ -20,24 +20,21 @@ import (
...
 	for _, container := range containers {
 		for _, key := range containerArr {
 			if strings.HasPrefix(container.Image, key) {
-				fmt.Println(container)
-				// stop appsody containers with "cw-" in their name, and all other cw containers
-				if key != "appsody" || strings.Contains(container.Names[0], "cw-") {
-					fmt.Println("Stopping container ", container.Names, "... ")
-					utils.StopContainer(container)
-				}
+				fmt.Println("Stopping container ", container.Names[0], "... ")
```
makandre: The check for the `cw-` container prefix (for appsody) was put in to avoid stopping appsody containers that are started outside of Codewind

utils/docker.go
```diff
@@ -291,9 +291,13 @@ func StopContainer(container types.Container) {
 		errors.CheckErr(err, 108, "")
 	}
 
-	// Remove the container so it isnt lingering in the background
-	if err := cli.ContainerRemove(ctx, container.ID, types.ContainerRemoveOptions{}); err != nil {
-		errors.CheckErr(err, 108, "")
+	// Do not attempt to remove appsody images as that happens automatically
+	// when an appsody container stops
+	if !strings.HasPrefix(container.Image, "appsody") {
```
makandre: Rather than hardcode appsody here, could we allow the operation to succeed even if it fails to remove container?  (There could be other future extensions that starts a container with `--rm` flag which self-removes a stopped container)

tobespc: we could but I would rather add that as a new enhancement rather than keep this one going

actions/remove.go
```diff
@@ -22,17 +22,16 @@ import (
...
 	imageArr[1] = "eclipse/codewind-performance"
 	imageArr[2] = "eclipse/codewind-initialize"
 	imageArr[3] = "cw-"
-	imageArr[4] = "appsody"
 	networkName := "codewind"
 
 	if tag != "" {
 		for i := 0; i < len(imageArr); i++ {
-			if i == 3 || i == 4 {
+			if i == 3 {
```
hhellyer: Can we just change this whole section to:
```
    imageArr := []string{
        "eclipse/codewind-pfe-amd64:" + tag,
        "eclipse/codewind-performance-amd64:" + tag,
        "eclipse/codewind-initialize-amd64:" + tag,
        "cw-",
    }
```
and initialise the array directly rather than initialising it then walking over most of the elements?

actions/stop-all.go
tobespc: ok, will add back in

actions/stop-all.go
```diff
@@ -20,24 +20,21 @@ import (
...
 //StopAllCommand to stop codewind and project containers
 func StopAllCommand() {
-	containerArr := [2]string{}
+	containerArr := [4]string{}
 	containerArr[0] = "codewind-pfe"
 	containerArr[1] = "codewind-performance"
-	containerArr[0] = "cw-"
-	containerArr[1] = "appsody"
+	containerArr[2] = "cw-"
+	containerArr[3] = "appsody"
```
hhellyer: Can we also move this to a simpler array declaration:
```
containerArr := []string{
"codewind-pfe", 
"codewind-performance",
"cw-",
"appsody",
}
```
Given the original bug was because we were assigning to indexes manually it would make sense.

actions/remove.go
tobespc: agreed

actions/stop-all.go
tobespc: agreed

PR #112, opened 2019-09-10
Statically compile the linux binary so it can run on Alpine.

hhellyer:
This adds the `CGO_ENABLED=0` flag to produce a binary that can run on Alpine linux as well as normal Linux.

I've tested on a local Linux box that this produces a binary that runs inside the theia container but I've marked this WIP so I can download the binary Jenkins builds for this PR and verify that is good as well before this is merged. This PR is ready for review.


hhellyer:
Build uploads are disabled on PR requests so I'll check when this is merged. I need to manually add it to codewind-vscode to get the plugin to pick it up.

PR #114, opened 2019-09-10
use appsody binary for init

laven26:


utils/appsody.go
```diff
@@ -13,13 +13,26 @@
...
 	"log"
+	"runtime"
 )
  
  // SuccessfullyCallAppsodyInit calls Appsody Init to initialise Appsody projects and returns a boolean to indicate success
  func SuccessfullyCallAppsodyInit(projectPath string) (bool, error) {
-	cmd := exec.Command("appsody", "init")
+	const GOOS string = runtime.GOOS
+	cwd, err := os.Getwd()
+	appsodyBinPath := "/appsody"
```
makandre: This probably should come after the error check

PR #119, opened 2019-09-13
Call commands defined by extensions

makandre:
For https://github.com/eclipse/codewind/issues/442

Also fixes https://github.com/eclipse/codewind/issues/445

### DIFF
--- a/actions/project.go
+++ b/actions/project.go
@@ -71,8 +71,8 @@ func DownloadTemplate(c *cli.Context) {
 	}
 }
 
-// testIsExtension tests if a project is an extension project and run associated commands as necessary
-func testIsExtension(projectPath string) (string, error) {
+// checkIsExtension checks if a project is an extension project and run associated commands as necessary
+func checkIsExtension(projectPath string) (string, error) {
 
 	extensions, err := apiroutes.GetExtensions()
 	if err != nil {
@@ -115,7 +115,7 @@ func ValidateProject(c *cli.Context) {
 		Language:  language,
 		BuildType: buildType,
 	}
-	extensionType, err := testIsExtension(projectPath)
+	extensionType, err := checkIsExtension(projectPath)
 	if extensionType != "" {
 		if err == nil {
 			validationResult = ProjectType{
--- a/actions/project.go
+++ b/actions/project.go
@@ -78,18 +78,18 @@ func ValidateProject(c *cli.Context) {
 	validationStatus := "success"
 	// result could be ProjectType or string, so define as an interface
 	var validationResult interface{}
-	language, buildType, isAppsody := utils.DetermineProjectInfo(projectPath)
+	language, buildType := utils.DetermineProjectInfo(projectPath)
 	validationResult = ProjectType{
 		Language: language,
 		BuildType: buildType,
 	}
-	if isAppsody {
-		validated, err := utils.SuccessfullyCallAppsodyInit(projectPath)
-		if !validated {
-			validationStatus = "failed"
-			validationResult = err.Error()
-		}
-	}
+	// if isAppsody {
+	// 	validated, err := utils.SuccessfullyCallAppsodyInit(projectPath)
+	// 	if !validated {
+	// 		validationStatus = "failed"
+	// 		validationResult = err.Error()
+	// 	}
+	// }
 
 	response := ValidationResponse{
 		Status: validationStatus,
--- a/actions/project.go
+++ b/actions/project.go
@@ -19,6 +19,7 @@ import (
 	"path"
 	"regexp"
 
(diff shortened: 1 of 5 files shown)


makandre:
@liamchampton or @tobespc could you review this?

actions/project.go
```diff
@@ -52,9 +53,9 @@ func DownloadTemplate(c *cli.Context) {
 
 	// Remove invalid characters from the string we will use
 	// as the project name in the template.
-	r := regexp.MustCompile("[^a-zA-Z0-9._-]");
```
liamchampton: Good spot!

actions/project.go
```diff
@@ -70,6 +71,36 @@ func DownloadTemplate(c *cli.Context) {
 	}
 }
 
+// testIsExtension tests if a project is an extension project and run associated commands as necessary
+func testIsExtension(projectPath string) (string, error) {
```
liamchampton: Could this function be called something else? Only suggesting this as when we unit test the functions begin `TestFooBar()` and this func name could be miss read / miss leading as it is not actually a test

actions/project.go
```diff
@@ -70,6 +71,36 @@ func DownloadTemplate(c *cli.Context) {
...
+
+	extensions, err := apiroutes.GetExtensions()
+	if err != nil {
+		return "", err
```
liamchampton: As the func `GetExtensions()` can return an error, maybe it would be good to display it here at the top level of the func call in this if block. To do this you could use `log.printf()` or `log.fatal()`. The reason i suggest using the log library in this case is because it prints to stderr by default.

utils/extensions.go
```diff
@@ -9,43 +9,45 @@
...
 	cmd.Stdout = output
 	cmd.Stderr = output
 	if err := cmd.Start(); err != nil { // after 'Start' the program is continued and script is executing in background
-		log.Println("There was a problem initializing the Appsody project: ", err, ". Project was not initialized.")
-		return false, err
+		log.Println("There was a problem running the command:", command.Command)
+		return err
 	}
-	log.Printf("Please wait while the Appsody project is initialized... %s \n", output.String())
+	log.Printf("Please wait while the project is initialized... %s", output.String())
```
liamchampton: This should be `fmt.Printf()` as this is currently going to stderr by default and not stdout

actions/project.go
makandre: It was actually the auto-linter that changed this :)

makandre:
@liamchampton I tested this change with https://github.com/makandre/codewind-appsody-extension/blob/ext_cmd/codewind.yaml

On mac, I built the installer and copied it to the location where VSCode plugin puts the installer (alongside appsody CLI).  Then I invoked it on an appsody project:

```
andrews-mbp:codewind-installer andrewmak$ /var/folders/c8/12yjw3yn40gcmw3mpzv0sj6c0000gn/T/codewind-installer project /opt/Repos/Tempest/codewind/codewind-workspace/knnode/
Please wait while the project is initialized... 2019/09/18 09:50:37 *
*
*

A new CLI update is available.
Please go to https://github.com/appsody/appsody/releases/latest and update from 0.4.3 --> 0.4.4.

*
*
*
Setting up the development environment
Running command: docker pull appsody/nodejs-express:0.2 
Running command: docker run --rm --entrypoint /bin/bash appsody/nodejs-express:0.2 -c find /project -type f -name .appsody-init.sh 
Successfully initialized Appsody project

{"status":"success","projectPath":"/opt/Repos/Tempest/codewind/codewind-workspace/knnode/","result":{"language":"nodejs","projectType":"appsodyExtension"}}
```

From that I'm able to see the `appsody init` was called and now the `projectType` returned is also correct.

@sghung did similar testing on Windows

makandre:
Also, for error case (PFE not running):

```
andrews-mbp:codewind-installer andrewmak$ /var/folders/c8/12yjw3yn40gcmw3mpzv0sj6c0000gn/T/codewind-installer project /opt/Repos/Tempest/codewind/codewind-workspace/knnode/
2019/09/18 10:14:18 There was a problem retrieving extensions data
{"status":"failed","projectPath":"/opt/Repos/Tempest/codewind/codewind-workspace/knnode/","result":"Get http:///api/v1/extensions: http: no Host in request URL"}
```

utils/extensions.go
makandre: So I needed to revert this back to `log`.  Because stdout needed to be json only because it is parsed by IDE

makandre:
Ported part of fix https://github.com/eclipse/codewind-installer/pull/126 to this PR for master

makandre:
@liamchampton any other changes required for this PR?

PR #128, opened 2019-09-26
Rename codewind-installer -> cwctl

liamchampton:
**Problem (https://github.com/eclipse/codewind/issues/255):** Since the installer is more of a CLI now, it was deemed necessary to rename this from `codewind-installer` -> `cwctl`.

**Solution**
- produce a binary with the prefix `cwtl` instead of `codewind-installer`
- rename `codewind-installer` -> `cwctl` in the appropriate places across the code base
- linter fixed some format errors
- rename `installer-docker-compose.yaml` -> `docker-compose.yaml`

**Testing**
```
 ✓ invoke install command - install latest with --json
 ✓ invoke status -j command - output = '{status:stopped,installed-versions:[latest]}'
 ✓ invoke start command - Start dockerhub images (latest)
 ✓ invoke stop-all command - Stop dockerhub images (latest)
 ✓ invoke remove command - remove all dockerhub images

5 tests, 0 failures
```
**Extra**
This is just MVP. See https://github.com/eclipse/codewind/issues/255 for what needs to be done next to complete this name change.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -32,7 +32,7 @@ spec:
     }
 
     environment {
-        CODE_DIRECTORY_FOR_GO = 'src/github.com/eclipse/cwctl'
+        CODE_DIRECTORY_FOR_GO = 'src/github.com/eclipse/codewind-installer'
         DEFAULT_WORKSPACE_DIR_FILE = 'temp_default_dir'
     }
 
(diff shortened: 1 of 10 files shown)


tetchel:
maybe we could name the docker-compose `codewind-docker-compose` or similar, in case the user already has a file with that very common name in `.`

PR #132, opened 2019-10-01
Add security commands

markcor11:
## Problem

- Configuration of a Keycloak container can be complicated and should be simplified. 
- IDEs should call Codewind APIs in a way that does not require them to manage the users credentials

## Solution

Introduce a new set of security commands to the Codewind CLI to : 

1. Create and manage a Keycloak realm for Codewind users in an existing Keycloak container
2. Create the Codewind client IDs and secrets needed for apps to authenticate
3. Create an initial developer user account for the application developer 
4. Provide a password change capability to change default passwords to something unique
5. Adds documentation explaining the use of the commands

## Commands added : 

```
   sectoken - Authenticate and obtain an access_token
   secrealm - Manage Realm configuration
   secclient - Manage client access configuration
   secuser  - Manage keycloak user account
```

## To follow : 

1. Tests
2. Keychain storage of user credentials
3. Enable Integration with Portal APIs

### DIFF
--- a/actions/security.go
+++ b/actions/security.go
@@ -68,6 +68,21 @@ func SecurityClientGet(c *cli.Context) {
 	os.Exit(0)
 }
 
+// SecurityClientGetSecret : Retrieve a client secret from Keycloak
+func SecurityClientGetSecret(c *cli.Context) {
+	registredClientSecret, err := security.SecClientGetSecret(c)
+	if err != nil {
+		fmt.Println(err.Error())
+		os.Exit(0)
+	}
+	if registredClientSecret != nil {
+		utils.PrettyPrintJSON(registredClientSecret)
+		os.Exit(0)
+	}
+	utils.PrettyPrintJSON(security.Result{Status: "Not found"})
+	os.Exit(0)
+}
+
 // SecurityUserCreate : Create a user in a Keycloak realm
 func SecurityUserCreate(c *cli.Context) {
 	err := security.SecUserCreate(c)
@@ -94,8 +109,8 @@ func SecurityUserGet(c *cli.Context) {
 	os.Exit(0)
 }
 
-// SecurityUserSetPW : Set a users password in Keycloak
-func SecurityUserSetPW(c *cli.Context) {
+// SecurityUserSetPassword : Set a users password in Keycloak
+func SecurityUserSetPassword(c *cli.Context) {
 	err := security.SecUserSetPW(c)
 	if err != nil {
 		fmt.Println(err.Error())
--- a/actions/security.go
+++ b/actions/security.go
@@ -55,13 +55,13 @@ func SecurityClientCreate(c *cli.Context) {
 
 // SecurityClientGet : Retrieve a client configuration from Keycloak
 func SecurityClientGet(c *cli.Context) {
-	registredClient, err := security.SecClientGet(c)
+	registeredClient, err := security.SecClientGet(c)
 	if err != nil {
 		fmt.Println(err.Error())
 		os.Exit(0)
 	}
-	if registredClient != nil {
-		utils.PrettyPrintJSON(registredClient)
+	if registeredClient != nil {
+		utils.PrettyPrintJSON(registeredClient)
 		os.Exit(0)
 	}
 	utils.PrettyPrintJSON(security.Result{Status: "Not found"})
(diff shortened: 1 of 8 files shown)


utils/security/user.go
```diff
@@ -0,0 +1,193 @@
...
+	// Authenticate if needed
+	if accesstoken == "" {
+		authToken, err := SecAuthenticate(c, KeycloakMasterRealm, KeycloakAdminClientID)
+		if err != nil || authToken == nil {
+			return nil, err
+		}
+		accesstoken = authToken.AccessToken
+	}
+
+	// Built REST request
```
jcockbain96: Is this one meant to be build?

utils/security/user.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+		accesstoken = authToken.AccessToken
+	}
+
+	registeredUser, secError := SecUserGet(c)
+	if secError != nil {
+		return secError
+	}
+
+	// Built REST request
```
jcockbain96: Build again?

actions/security.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		fmt.Println(err.Error())
+		os.Exit(0)
+	}
+	if registeredClient != nil {
+		utils.PrettyPrintJSON(registeredClient)
+		os.Exit(0)
+	}
+	utils.PrettyPrintJSON(security.Result{Status: "Not found"})
+	os.Exit(0)
```
hhellyer: os.Exit(0) is probably the wrong way to end the commands.
- log.Fatal(msg) seems to be what we do elsewhere.
- The editor plugins look at the return codes to see if there's a problem. Having os.Exit(0) on an error path is probably going to cause them problems.

utils/security/authenticate.go
```diff
@@ -0,0 +1,83 @@
...
+	SessionState    string `json:"session_state"`
+	Scope           string `json:"scope"`
+}
+
+// SecAuthenticate - sends credentials to the auth server for a specific realm and returns an AuthToken
+// connectionRealm can be used to override the supplied context arguments
+func SecAuthenticate(c *cli.Context, connectionRealm string, connectionClient string) (*AuthToken, *SecError) {
+	if c.GlobalBool("insecure") {
+		http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+	}
```
hhellyer: If lots of places are going to read the insecure flag I think we can add it to app.Before() in commands.go.
(Disclaimer - I had a branch for uploading that needed to do that and had that change in.)

hhellyer:
```go
app.Before = func(c *cli.Context) error {
        // Handle Global flag to disable certificate checking here rather than in each command.
        if c.GlobalBool("insecure") {
                http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
        }
        return nil
}
```

utils/security/client.go
```diff
@@ -0,0 +1,216 @@
...
+
+	// build REST request
+	url := hostname + "/auth/admin/realms/" + realm + "/clients?clientId=" + clientid
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, &SecError{errOpConnection, err, err.Error()}
+	}
+	req.Header.Add("Authorization", "Bearer "+accesstoken)
+	req.Header.Add("cache-control", "no-cache")
+	req.Header.Add("cache-control", "no-cache")
```
hhellyer: The `cache-control` header is added twice. Above you add it in upper and lower case. Did you mean to do that here?

utils/security/user.go
```diff
@@ -0,0 +1,217 @@
...
+	if c.GlobalBool("insecure") {
+		http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+	}
+	hostname := strings.TrimSpace(strings.ToLower(c.String("host")))
+	realm := strings.TrimSpace(c.String("realm"))
+	accesstoken := strings.TrimSpace(c.String("accesstoken"))
+	newPassword := strings.TrimSpace(c.String("newpw"))
+
+	// validate password characters
+	if strings.Contains(newPassword, "'") || strings.Contains(newPassword, "\"") {
```
hhellyer: Is it just quotes that are banned from passwords? Is that a keycloak limitation?

actions/security.go
markcor11: I'll check with the IDE folks to see what they are expecting. My thought process was,  the Exit(0) is that the CLI command completed correctly with no fault.  However, that does not mean the result of the command was successful if the backend server the CLI is talking to returns an error message.  In that last case the CLI would return a JSON response with the addition of {error: "operation", "error-description": "some meaningful text"}.   That message is still output and returned by the CLI for a human or the IDE to read.

utils/security/client.go
markcor11: thanks - resolved

utils/security/user.go
markcor11: I was a bit nervous about allowing quotes in passwords to avoid any injection or invalid escape sequences being posted in JSON.

utils/security/authenticate.go
markcor11: Just tried that and it looks good,  made changes and remove the duplicates

utils/security/user.go
markcor11: I tried a few combinations with double and single quotes which did marshal ok,  and I was able to login with a changed password through the CLI and though a browser so I have removed this check.

PR #133, opened 2019-10-01
Fix missed variable rename

liamchampton:
**Problem**
Master build failed due to missing rename

**Solution**
Fix the rename


hhellyer:
Merging ahead of the PR build as there's no point waiting, the PR build won't run the broken step.

PR #147, opened 2019-10-11
Use `DockerError` Struct in `utils/docker`

jcockbain:
Resolves https://github.com/eclipse/codewind/issues/462 by returning a message for if a user is calling start without the correctly tagged images. 

Introduces a `DockerError` struct to handle all errors returned from `utils/docker`, in the style of the errors added to `deployment` and `security` utils. 

Checking is added in the commands that call these functions to return these errors.


jcockbain:
WIP, as testing each relevant command having rebased.

PR #162, opened 2019-10-23
Don't add some directories and files into the PFE container to save time

james-wallis:
PR to add the functionality discussed in [Issue #750](https://github.com/eclipse/codewind/issues/750).

Has two separate lists for files and directories.


utils/project/sync.go
```diff
@@ -143,3 +154,56 @@ func completeUpload(projectId string, files []string, modfiles []string, timesta
 		// TODO - Need to handle this gracefully.
 	}
 }
+
+func ignoreFileOrDirectory(name string, isDir bool) bool {
+	// List of files not add to PFE
```
liamchampton: typo: should read as "List of files not added to PFE" or something similar. Current comment doesn't make sense.

utils/project/sync.go
```diff
@@ -143,3 +154,56 @@ func completeUpload(projectId string, files []string, modfiles []string, timesta
...
+		".bluemix",
+		"terraform",
+		".build-ubuntu"}
+
+	ignoredList := ignoredFiles
+	if isDir {
+		ignoredList = ignoredDirectories
+	}
+
+	fileInIgnored := false
```
liamchampton: Can this variable be called something more explicit to make it easier to read? fileInIgnored..?

utils/project/sync.go
james-wallis: Changed to "List of directories that will not be sent to PFE"

utils/project/sync.go
james-wallis: Changed to "isFileInIgnoredList"

james-wallis:
@liamchampton I've added unit tests for this function.

utils/project/sync_test.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			fileIsIgnored := ignoreFileOrDirectory(test.name, test.isDir)
+
+			assert.IsType(t, test.shouldBeIgnored, fileIsIgnored, "Got: %s", fileIsIgnored)
+
+			assert.Equal(t, test.shouldBeIgnored, fileIsIgnored, "fileIsIgnored was %b but should have been %b", fileIsIgnored, test.shouldBeIgnored)
+		})
+		fmt.Println()
```
liamchampton: this println is not needed

utils/project/sync.go
```diff
@@ -143,3 +154,56 @@ func completeUpload(projectId string, files []string, modfiles []string, timesta
...
+		".cfignore",
+		"localm2cache.zip",
+		"libertyrepocache.zip",
+		"run-dev",
+		"run-debug",
+		"manifest.yml",
+		"idt.js",
+		".bluemix",
+		".build-ubuntu",
+		".yo-rc.json"}
```
liamchampton: bracket on new line and finish with a trailing comma

utils/project/sync.go
```diff
@@ -143,3 +154,56 @@ func completeUpload(projectId string, files []string, modfiles []string, timesta
...
+		"load-test*",
+		".settings",
+		"Dockerfile-tools",
+		"target",
+		"mc-target",
+		".m2",
+		"debian",
+		".bluemix",
+		"terraform",
+		".build-ubuntu"}
```
liamchampton: same as above

jgwest:
@james-wallis I presume the hardcoded ignore list is temporary until a more elegant solution is in plac?

PR #165, opened 2019-10-25
Use `depID` for per project deployments

jcockbain96:
### Project Bind additions ###

- Takes an extra flag, depID, on the `bind` command. This is used as the API host to call `/api/v1/bind/start`. 

- Generates a `HOME/.codewind/config/deployments/{projectID}.json` file, which maps each projectID (unique to each deployment) to its `depId`.

- Then reads that file, and uses its `depID` to find its corresponding deployment URL. This is then used as the host for API calls on that project. 

### Extra ###

- Documents `bind` and `sync` in the README.md. 

- Fixes go-lint warnings in `bind` and `sync` files. 

### Tests ###

- Can manually bind a project to local deployment (existing creation). Can call bind on a project with a non-local deployment ID, and then call its deployment host, added to `deployments.json`.

- Adds a unit test for `getDeploymentURL` when there are no deployments in `deployments.json`. Adding a test for when there are deployments now.

We will change `deployments` to `connections` in a future PR, and add further testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,22 @@ $ brew upgrade dep
 
 Subcommands:</br>
 
+`bind` - Bind a project to Codewind for building and running
+
+> **Flags:**
+> --name,n value                Project name
+> --language,l value            Project language
+> --type,t value                Project Type
+> --path,p value                Project Path
+> --deployment ID,d value       Deployment ID
+
+`sync` - Synchronize a bound projects files, to its deployment
+
+> **Flags:**
+> --path,l value                Project Path
+> --id, i value                 Project ID
+> --time, t value               Time of last sync for Project
+
 `deployment/dep` - Manage the deployment targets for a project
 
 `add,a` - Add a deployment to a project
@@ -165,7 +181,7 @@ Subcommands:</br>
 
 Subcommands:</br>
 
-`get/g` - Authenticate and obtain an access_token. 
+`get/g` - Authenticate and obtain an access_token.
 
 >**Note 1:**: The preferred way to authenticate is by supplying just the depid and username. In this mode the command will use the stored password from the platform keyring
 >**Note 2:**: If you dont have a depid you must supply use the host, realm and client flags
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -109,6 +109,22 @@ $ brew upgrade dep
...
 Subcommands:</br>
 
+`bind` - Bind a project to Codewind for building and running
+
+> **Flags:**
+> --name,n value                Project name
+> --language,l value            Project language
+> --type,t value                Project Type
+> --path,p value                Project Path
+> --deployment ID,d value       Deployment ID
```
markcor11: Should this be depid,d ?

actions/commands.go
```diff
@@ -76,6 +76,7 @@ func Commands() {
 						cli.StringFlag{Name: "language, l", Usage: "the project language", Required: true},
 						cli.StringFlag{Name: "type, t", Usage: "the type of the project", Required: true},
 						cli.StringFlag{Name: "path, p", Usage: "the path to the project", Required: true},
+						cli.StringFlag{Name: "depID, d", Usage: "the deployment id for the project", Required: true},
```
markcor11: I think we're currently using depid,d throughout the other commands

utils/project/sync.go
```diff
@@ -129,14 +147,14 @@ func syncFiles(projectPath string, projectId string, synctime int64) ([]string,
...
 
 	payload := &CompleteRequest{FileList: files, ModifiedList: modfiles, TimeStamp: timestamp}
 	jsonPayload, _ := json.Marshal(payload)
 
 	// Make the request to end the sync process.
-	resp, err := http.Post(uploadEndUrl, "application/json", bytes.NewBuffer(jsonPayload))
+	resp, err := http.Post(uploadEndURL, "application/json", bytes.NewBuffer(jsonPayload))
 	fmt.Println("Upload end status:" + resp.Status)
 	if err != nil {
 		panic(err)
```
markcor11: Not strictly part of this PR and I do see the TODO so until we figure out a way to undo/retry this seem ok. Can the IDE handle the panic message ?

PR #169, opened 2019-10-28
Remove `target` from projectConnection commands

jcockbain96:
## Summary

- Removes references to `target` in the code for projectConnections. 

- Updates commands to fit with the 1:1 projectID to connection mapping. `set`, `get`, and `remove`. 

- Simplifies the format of `.codewind/config/{projectID}.json`, now that we won't have multiple connections for a projectID.

``` 
{
	"schemaVersion": 1,
	"connectionID": "local"
}
```

## Testing

- Updated `connectionDeployment_test.js` for modified function.

- Manual testing of binding a project and getting, setting and removing its connectionID.


jcockbain96:
@tetchel sure, will do

jcockbain96:
Have pushed those changes. Since the `status` command was calling that `getTargetConnection` function, I have added a --conid flag to `status`. If this is set it will check for the remote connection status, if not it will check local.

PR #172, opened 2019-10-29
remove codewind project containers and networks before start

yangcao77:
Signed-off-by: Stephanie <stephanie.cao@ibm.com>

For issue https://github.com/eclipse/codewind/issues/611

If the `codewind_network` has already existed, but with a different `com.docker.network.bridge.host_binding_ipv4`. It needs to be recreated to apply the change.

All project containers need to be stopped before removing the `codewind_network`.

Run `stop-all` to stop all running codewind containers and remove `codewind_network` to fix this issue.


liamchampton:
Update: Spoken to @yangcao77 about this and since this is potentially just a one off issue due to an IP miss-match, this shouldn't be much of an issue to leave like this and find a more elegant way later on if it becomes a common problem

PR #175, opened 2019-10-30
Remove bind mount to ensure PFE can start on Windows.

hhellyer:
This PR disables bind mounts on Windows to allow pfe to start, it also moves docker-compose code to docker.go.

This is for  https://github.com/eclipse/codewind/issues/910

Signed-off-by: Howard Hellyer <hhellyer@uk.ibm.com>

### DIFF
--- a/utils/docker.go
+++ b/utils/docker.go
@@ -29,6 +29,7 @@ import (
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/docker/docker/pkg/term"
 	"github.com/eclipse/codewind-installer/errors"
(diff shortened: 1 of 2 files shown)


hhellyer:
@makandre - Could you take a look at this PR ( and eclipse/codewind#910 ) as this is a problem on Windows machines.

makandre:
I've commented on https://github.com/eclipse/codewind/issues/910.  If we remove this bind mount, then Appsody projects would not work under Windows

We had `C:\codewind-workspace` mounted on Windows pre 0.6.0, so I'm not sure why this bind mount is causing trouble

PR #177, opened 2019-10-30
Use the ignoredPaths in .cw-settings to ignoreFiles which will be added into Portal

james-wallis:
### Summary
I've added functionality to take a users `ignoredPaths` which can be defined in `.cw-settings` and add it to the hard coded list of ignored files. 
Addresses the issue raised in https://github.com/eclipse/codewind/issues/903 and but @elsony in https://github.com/eclipse/codewind/issues/835.
This PR doesn't do any work to remove the hard coded list.

### Additional
I've moved what was in `utils/projects.go` into `utils/project/create.go` as the functions in projects.go were called only in create.go. This allows the new code to use the `CWSettings` struct in `create.go` (was in `projects.go`). I've also changed some functions that are only called locally to have a lowercase letter at the start to show they're only called locally.

The tests that were in `utils/projects_test.go` are now `utils/project/create_test.go`.

### Testing
I've added unit tests for the new `retrieveIgnoredPathsListFromCWSettings` function and improved other unit tests which call this function.

I've run the unit tests in `utils/project` to ensure that they all still pass.


utils/project/sync.go
```diff
@@ -196,7 +198,21 @@ func completeUpload(projectID string, files []string, modfiles []string, conURL
 	return resp.Status, resp.StatusCode
 }
 
-func ignoreFileOrDirectory(name string, isDir bool) bool {
+// Retrieve the ignoredPaths list from a .cw-settings file
+func retrieveIgnoredPathsListFromCWSettings(projectPath string) []string {
```
liamchampton: This func name is rather long. Can this be shortened?

utils/project/sync_test.go
```diff
@@ -12,70 +12,188 @@
...
+	// Create .cw-settings files
+	file, _ := json.Marshal(populatedIgnoredPaths)
+	ioutil.WriteFile(cwSettingsPopulatedPath+"/.cw-settings", file, 0644)
+	file, _ = json.Marshal(emptyIgnoredPaths)
+	ioutil.WriteFile(cwSettingsEmptyPath+"/.cw-settings", file, 0644)
+	file, _ = json.Marshal(noIgnoredPaths)
+	ioutil.WriteFile(cwSettingsNoIgnoredPaths+"/.cw-settings", file, 0644)
+}
+
+func shutdown() {
```
liamchampton: this func name doesnt tell me it is removing a test folder. I would suggest using `removeTestDir()` or something similar

utils/project/sync_test.go
```diff
@@ -12,70 +12,188 @@
...
+			fileIsIgnored := ignoreFileOrDirectory(test.name, test.isDir, test.ignoredPathsList)
 
 			assert.IsType(t, test.shouldBeIgnored, fileIsIgnored, "Got: %s", fileIsIgnored)
 
 			assert.Equal(t, test.shouldBeIgnored, fileIsIgnored, "fileIsIgnored was %b but should have been %b", fileIsIgnored, test.shouldBeIgnored)
 		})
 	}
 }
+
+func TestRetrieveIgnoredPathsListFromCWSettings(t *testing.T) {
```
liamchampton: If you change the func name above, this will also change too

utils/project/sync_test.go
james-wallis: have removed this function

utils/project/sync_test.go
james-wallis: Changed

utils/project/sync.go
james-wallis: Made it smaller

PR #192, opened 2019-11-06
CLI Che Sync Fix 

ebuckle:
- Moving connection and projects dirs to use the shared workspace between Theia and sidecar when on Che
- Performs a check to see if Che environment variables are set to see if the CLI is being run in Che before setting the home dir
- Allows connection data to be shared between the two copies of cwctl, fixing [issue 906](https://github.com/eclipse/codewind/issues/906)

Signed-off-by: Edward Buckle <edward.buckle0@gmail.com>

### DIFF
--- a/pkg/utils/connections/connection.go
+++ b/pkg/utils/connections/connection.go
@@ -239,12 +239,23 @@ func saveConnectionsConfigFile(ConnectionConfig *ConnectionConfig) *ConError {
 
 // getConnectionConfigDir : get directory path to the connections file
 func getConnectionConfigDir() string {
-	const GOOS string = runtime.GOOS
+	val, isSet := os.LookupEnv("CHE_API_EXTERNAL")
 	homeDir := ""
-	if GOOS == "windows" {
-		homeDir = os.Getenv("USERPROFILE")
+	if isSet && (val != "") {
+		val, isSet := os.LookupEnv("CHE_PROJECTS_ROOT")
+		if isSet && (val != "") {
+			homeDir = val
+		} else {
+			// Cannot set projects root without env variable, suggests issue with Codewind Che installation
+			panic("CHE_PROJECTS_ROOT not set")
+		}
 	} else {
-		homeDir = os.Getenv("HOME")
+		const GOOS string = runtime.GOOS
+		if GOOS == "windows" {
+			homeDir = os.Getenv("USERPROFILE")
+		} else {
+			homeDir = os.Getenv("HOME")
+		}
 	}
 	return path.Join(homeDir, ".codewind", "config")
 }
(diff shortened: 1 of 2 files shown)


pkg/utils/connections/connection.go
```diff
@@ -239,12 +239,22 @@ func saveConnectionsConfigFile(ConnectionConfig *ConnectionConfig) *ConError {
 
 // getConnectionConfigDir : get directory path to the connections file
 func getConnectionConfigDir() string {
-	const GOOS string = runtime.GOOS
+	val, ok := os.LookupEnv("CHE_API_EXTERNAL")
```
liamchampton: what is `ok`? If this is a bool, can this var name be conveyed in a better way?

pkg/utils/connections/connection.go
```diff
@@ -239,12 +239,22 @@ func saveConnectionsConfigFile(ConnectionConfig *ConnectionConfig) *ConError {
...
 	homeDir := ""
-	if GOOS == "windows" {
-		homeDir = os.Getenv("USERPROFILE")
+	if ok && (val != "") {
+		val, ok := os.LookupEnv("CHE_PROJECTS_ROOT")
+		if ok && (val != "") {
+			homeDir = val
+		} else {
+			panic("CHE_PROJECTS_ROOT not set")
```
liamchampton: We don't usually want to use panics but this is necessary in this case. Please can this have a comment to explain and prevent it being removed by accident in future?

pkg/utils/project/projectDeployment.go
```diff
@@ -158,12 +158,22 @@ func RemoveConnectionFile(projectID string) *ProjectError {
 
 // getProjectConnectionConfigDir : Get directory path to the connection file
 func getProjectConnectionConfigDir() string {
-	const GOOS string = runtime.GOOS
+	val, ok := os.LookupEnv("CHE_API_EXTERNAL")
```
liamchampton: same as above

pkg/utils/project/projectDeployment.go
```diff
@@ -158,12 +158,22 @@ func RemoveConnectionFile(projectID string) *ProjectError {
...
 	homeDir := ""
-	if GOOS == "windows" {
-		homeDir = os.Getenv("USERPROFILE")
+	if ok && (val != "") {
+		val, ok := os.LookupEnv("CHE_PROJECTS_ROOT")
+		if ok && (val != "") {
+			homeDir = val
+		} else {
+			panic("CHE_PROJECTS_ROOT not set")
```
liamchampton: same as above

pkg/utils/connections/connection.go
ebuckle: ok is a bool for whether the env variable is set or not. val is then the value of the env variable.

Changing to isSet for better clarity.

pkg/utils/connections/connection.go
ebuckle: Short explanation that an unset env variable suggests a wider issue with Che install of codewind

PR #202, opened 2019-11-11
JSON return for UpgradeProjects

ebuckle:
For [#961](https://github.com/eclipse/codewind/issues/961)

Removed prints from UpgradeProjects, replaced with second return value containing a breakdown of which projects succeeded and which failed to upgrade. 

This enables UpgradeProjects to exit with code 1 if any projects fail to upgrade and allows the JSON formatted output be parsed.

Signed-off-by: Edward Buckle <edward.buckle0@gmail.com>

### DIFF
--- a/pkg/utils/project/upgrade.go
+++ b/pkg/utils/project/upgrade.go
@@ -14,7 +14,6 @@ package project
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


pkg/utils/project/upgrade.go
```diff
@@ -59,28 +59,29 @@ func UpgradeProjects(c *cli.Context) *ProjectError {
 			projectType := result["projectType"]
 			name := result["name"]
 			location := oldDir + "/" + name
-			fmt.Println("Calling bind for project " + name + "," + projectType + "," + language + " in " + location)
 
 			if language != "" && projectType != "" && name != "" && location != "" {
-				response, binderr := Bind(location, name, language, projectType, "local")
-				PrintAsJSON := c.GlobalBool("json")
+				_, binderr := Bind(location, name, language, projectType, "local")
```
liamchampton: `binderr` -> `bindErr` (camel case for vars to keep consistent)

liamchampton:
Is there any tests for this? / how has this been tested?

ebuckle:
Isn't currently a test file for upgrade.go, definitely something that needs adding.

New output: 
Screen Shot 2019-11-11 at 14 34 09

liamchampton:
I think this should have some unit test written to accompany this change. Some short and simple ones could include checking for a workspace, checking for a project inf file and checking what it can extract from it.

ebuckle:
Sounds good, I'll get that added in

PR #208, opened 2019-11-13
Enable auth on api routes (Bind and Sync)

markcor11:
## Problem

Enable authentication on Bind and Sync Issue https://github.com/eclipse/codewind/issues/900

## Solution

* Reorganize code layout to resolve the circular import errors seen when using code across packages
* Adds a Auth wrapper around PFE API requests (Bind and Sync)
* Adds a new global flag "loglevel" to produce extended debug output when functions adopt new logger


pkg/project/bind.go
```diff
@@ -152,15 +155,18 @@ func Bind(projectPath string, name string, language string, projectType string,
...
 
 	// Make the request to end the sync process.
-	resp, err := http.Post(uploadEndURL, "application/json", bytes.NewBuffer(jsonPayload))
-	if err != nil {
+	request, err := http.NewRequest("POST", uploadEndURL, bytes.NewBuffer(jsonPayload))
+	request.Header.Set("Content-Type", "application/json")
+	resp, httpSecError := sechttp.DispatchHTTPRequest(http.DefaultClient, request, username, connection.ID)
+
+	if httpSecError != nil {
 		panic(err)
```
liamchampton: Shouldn't being using a panic to catch the error as it throws the whole stack. Not part of this PR and out of scope for a change but needs to be addressed.

pkg/actions/commands.go
```diff
@@ -95,6 +101,7 @@ func Commands() {
 						cli.StringFlag{Name: "path, p", Usage: "the path to the project", Required: true},
 						cli.StringFlag{Name: "id, i", Usage: "the project id", Required: true},
 						cli.StringFlag{Name: "time, t", Usage: "time of the last sync for the given project", Required: true},
+						cli.StringFlag{Name: "username,u", Usage: "Account Username", Required: false},
```
jcockbain96: could u be confused with url?

jcockbain96: Also, `Account Username` not in keeping with format of other Usage statements.

pkg/sechttp/httpclient.go
```diff
@@ -29,21 +29,19 @@ import (
 // Returns: HTTPResponse, HTTPSecError
 func DispatchHTTPRequest(httpClient utils.HTTPClient, originalRequest *http.Request, username string, connectionID string) (*http.Response, *HTTPSecError) {
 
-	logr.SetLevel(logr.TraceLevel)
-
-	logr.Printf("Request URL: %v %v", originalRequest.Method, originalRequest.URL)
+	logr.Tracef("Request URL: %v %v\n", originalRequest.Method, originalRequest.URL)
```
liamchampton: Good to leave these in since we can now change the level of logging. Useful information for developers.

pkg/actions/commands.go
markcor11: * CWCTL already uses username and its alias "u" in other a few other commands
* For the URL flag we seem to have standardised on "url" without an alias apart from 
project create which has both

pkg/project/bind.go
```diff
@@ -54,22 +55,25 @@ type (
 	}
 )
 
+// BindProject : Bind a project
 func BindProject(c *cli.Context) (*BindResponse, *ProjectError) {
 	projectPath := strings.TrimSpace(c.String("path"))
 	Name := strings.TrimSpace(c.String("name"))
 	Language := strings.TrimSpace(c.String("language"))
 	BuildType := strings.TrimSpace(c.String("type"))
+	cliUsername := strings.TrimSpace(strings.ToLower(c.String("username")))
```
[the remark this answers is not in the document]
markcor11: yes - I like camel case :-)

pkg/project/bind.go
markcor11: Agreed - should be addressed in a new PR

pkg/actions/commands.go
```diff
@@ -659,6 +659,17 @@ func Commands() {
...
+			case loglevel == "fatal":
+				logr.SetLevel(logr.FatalLevel)
+				break
+			case loglevel == "error":
+				logr.SetLevel(logr.ErrorLevel)
+				break
+			default:
+				logr.SetLevel(logr.InfoLevel)
 			}
 		} else {
```
liamchampton: If you set the `value` of the flag to Info level for logging, you _shouldnt_ need this `if != "" / else` clause as it will take that as the default if the flag is not mentioned

PR #209, opened 2019-11-13
Add logrus for CLI output

liamchampton:
**Problem https://github.com/eclipse/codewind/issues/937**
We currently have mixed output that needs to be standardized.

**Solution**
Making a start on standardizing the output the CLI produces. This PR introduces the `logrus` package. This will allow us to have more control over what the CLI outputs and at what level. **This PR concentrates on non-json output for commands**, replacing `fmt.Println()` with `logr.Infoln() where appropriate. It looks very similar to the logs within of PFE.

User Friendly Terminal Output:
```
INFO[0000] Please wait whilst containers initialize...   
INFO[0003] Creating network "codewind_network" with the default driver
Creating codewind-performance ... done
Creating codewind-pfe         ... done
 
INFO[0003] Waiting for Codewind to start                
..
INFO[0006] HTTP Response Status:200OK                   
INFO[0006] Codewind successfully started on http://127.0.0.1:10000 
```
Under the covers this output will look like:
```
time="2019-11-13T11:46:01Z" level=info msg="Waiting for Codewind to start"
..
time="2019-11-13T11:46:04Z" level=info msg="HTTP Response Status:200OK"
time="2019-11-13T11:46:04Z" level=info msg="Codewind successfully started on http://127.0.0.1:10000"
```

**Testing**
BATS test result:
`18 tests, 0 failures, 6 skipped`

**Extra** 
Added logic to prevent a redundant output line being printed by the new logger. If any unexpected output is present after running the image tagging exec command it should still be caught and printed. See link below to skip to the line within this PR.

https://github.com/eclipse/codewind-installer/compare/master...liamchampton:add-logrus?expand=1#diff-02a4c9edab65df5dfdd065f5b995d1faR191

### DIFF
--- a/pkg/utils/docker.go
+++ b/pkg/utils/docker.go
@@ -238,9 +238,11 @@ func ValidateImageDigest(image string) (string, *DockerError) {
 func TagImage(source, tag string) {
 	out, err := exec.Command("docker", "tag", source, tag).Output()
 	errors.CheckErr(err, 102, "Image Tagging Failed")
-
-	output := string(out[:])
-	logr.Infoln(output)
+	if len(out) > 0 {
+		// if 'out' byte array is ever > 0 then print the contents as string
+		output := string(out[:])
+		logr.Infoln(output)
+	}
 }
 
 // CheckContainerStatus of Codewind running/stopped
--- a/pkg/utils/docker.go
+++ b/pkg/utils/docker.go
@@ -14,11 +14,13 @@ package utils
 import (
 	"bytes"
 	"context"
+<<<<<<< HEAD
 	"encoding/json"
 	goErr "errors"
 	"fmt"
+=======
+>>>>>>> replace 'fmt.Print' with logrus #2
 	"io"
-	"log"
 	"net"
 	"os"
 	"os/exec"
@@ -113,8 +115,8 @@ func DockerCompose(tempFilePath string, tag string) {
 
 	const GOARCH string = runtime.GOARCH
 	const GOOS string = runtime.GOOS
-	fmt.Println("System architecture is: ", GOARCH)
-	fmt.Println("Host operating system is: ", GOOS)
+	logr.Infoln("System architecture is: ", GOARCH)
+	logr.Infoln("Host operating system is: ", GOOS)
 
 	if GOARCH == "x86_64" || GOARCH == "amd64" {
 		os.Setenv("PLATFORM", "-amd64")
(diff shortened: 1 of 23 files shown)


pkg/utils/docker.go
```diff
@@ -189,9 +188,11 @@ func PullImage(image string, jsonOutput bool) {
...
 	out, err := exec.Command("docker", "tag", source, tag).Output()
 	errors.CheckErr(err, 102, "Image Tagging Failed")
-
-	output := string(out[:])
-	fmt.Println(output)
+	if len(out) > 0 {
+		// if 'out' byte array is ever > 0 then print the contents as string
+		output := string(out[:])
+		logr.Infoln(output)
+	}
```
markcor11: Is this to avoid a blank line in the log output ?

pkg/utils/filesystem.go
```diff
@@ -71,9 +72,9 @@ func WriteToComposeFile(tempFilePath string, debug bool) bool {
 	errors.CheckErr(err, 203, "")
 
 	if debug == true {
-		fmt.Printf("==> %s structure is: \n%s\n\n", tempFilePath, string(marshalledData))
+		logr.Infof("==> %s structure is: \n%s\n\n", tempFilePath, string(marshalledData))
 	} else {
-		fmt.Println("==> environment structure written to " + tempFilePath)
+		logr.Infoln("==> environment structure written to " + tempFilePath)
```
markcor11: There's a couple of places throughout where the logging uses the formatter vs string concatenation.  Here's an example where both are used.  Nothing wrong with it,  but might be worth standardising.  I personally prefer the formatted version.

pkg/utils/filesystem.go
```diff
@@ -98,16 +99,18 @@ func DeleteTempFile(filePath string) (bool, error) {
...
 	for i := 0; i < 120; i++ {
 		resp, err := http.Get("http://" + hostname + ":" + port + healthEndpoint)
+		// following prints have been left in to prevent awkward terminal output
 		if err != nil {
 			fmt.Printf(".")
 		} else {
 			if resp.StatusCode == 200 {
-				fmt.Println("\nHTTP Response Status:", resp.StatusCode, http.StatusText(resp.StatusCode))
-				fmt.Println("Codewind successfully started on http://" + hostname + ":" + port)
+				fmt.Print("\n") // using fmt to ass a new line to separate from '...' in the output
```
markcor11: ass ?

pkg/utils/docker.go
liamchampton: Yes. A blank line looked weird when it is just `INFO[0001]`. With this logic it will only print output if there is any

pkg/utils/filesystem.go
liamchampton: Sure thing, i totally agree.

pkg/utils/filesystem.go
liamchampton: lol, oops.

PR #212, opened 2019-11-14
Set the minimum Docker API version to 1.30

johnmcollier:
Fixes https://github.com/eclipse/codewind/issues/731

This PR sets the minimum Docker API version for `cwctl` to 1.30. This version corresponds to Docker 17.06, which is the same minimum Docker version we had back in Microclimate.


johnmcollier:
As a note - it would probably be easier if we initialized the Docker API and context in one location, and had all of these functions access it.

PR #217, opened 2019-11-14
Add connection ID to every command which makes a request to PFE.

james-wallis:
### Summary 
* Added connection ID to templates, extensions and ignored-paths API calls.
* Added a `--conid` flag to the` template` and `project create`.
* Moved gatekeeper into its own directory to remove circular dependency.
* Moved config into the pkg directory to remove circular dependency.

### Testing
* Manual testing to check that templates cwctl commands still work with the change.
* I've added a single test for config which checks that it fails correctly when the connection ID does not exist. Only one test added as for any "local" connection ID tests, PFE is required to be running.
* Also updated tests to pass through a connection ID as "local"

### DIFF
--- a/pkg/actions/commands.go
+++ b/pkg/actions/commands.go
@@ -86,7 +86,7 @@ func Commands() {
 						cli.StringFlag{Name: "language, l", Usage: "The project language", Required: true},
 						cli.StringFlag{Name: "type, t", Usage: "The type of the project", Required: true},
 						cli.StringFlag{Name: "path, p", Usage: "The path to the project", Required: true},
-						cli.StringFlag{Name: "conid", Usage: "The connection id for the project", Required: false},
+						cli.StringFlag{Name: "conid", Value: "local", Usage: "The connection id for the project", Required: false},
 						cli.StringFlag{Name: "username,u", Usage: "Account Username", Required: false},
 					},
 					Action: func(c *cli.Context) error {
@@ -120,7 +120,8 @@ func Commands() {
 							Usage:   "Set connectionID for a project",
 							Flags: []cli.Flag{
 								cli.StringFlag{Name: "id,i", Usage: "Project ID", Required: true},
-								cli.StringFlag{Name: "conid", Usage: "Connection ID", Required: true},
+								// TODO
+								cli.StringFlag{Name: "conid", Value: "local", Usage: "Connection ID", Required: false},
 							},
 							Action: func(c *cli.Context) error {
 								ProjectSetConnection(c)
@@ -291,6 +292,12 @@ func Commands() {
 							Name:  "showEnabledOnly",
 							Usage: "Filter by whether a template is enabled or not",
 						},
+						cli.StringFlag{
+							Name:     "conid",
+							Value:    "local",
+							Usage:    "Connection ID",
+							Required: false,
+						},
 					},
 					Action: func(c *cli.Context) error {
 						ListTemplates(c)
@@ -300,8 +307,16 @@ func Commands() {
 				{
 					Name:  "styles",
 					Usage: "List available template styles",
+					Flags: []cli.Flag{
+						cli.StringFlag{
+							Name:     "conid",
+							Value:    "local",
+							Usage:    "Connection ID",
+							Required: false,
+						},
+					},
 					Action: func(c *cli.Context) error {
-						ListTemplateStyles()
+						ListTemplateStyles(c)
 						return nil
 					},
 				},
@@ -313,8 +328,16 @@ func Commands() {
 							Name:    "list",
 							Aliases: []string{"ls"},
 							Usage:   "List available template repos",
+							Flags: []cli.Flag{
+								cli.StringFlag{
+									Name:     "conid",
+									Value:    "local",
+									Usage:    "Connection ID",
+									Required: false,
+								},
+							},
 							Action: func(c *cli.Context) error {
-								ListTemplateRepos()
+								ListTemplateRepos(c)
 								return nil
 							},
 						},
@@ -336,6 +359,12 @@ func Commands() {
 									Value: "",
 									Usage: "Name of the template repo",
 								},
+								cli.StringFlag{
+									Name:     "conid",
+									Value:    "local",
+									Usage:    "Connection ID",
+									Required: false,
+								},
 							},
 							Action: func(c *cli.Context) error {
 								AddTemplateRepo(c)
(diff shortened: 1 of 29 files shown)


pkg/apiroutes/templates.go
```diff
@@ -54,8 +54,13 @@ type (
 
 // GetTemplates gets project templates from PFE's REST API.
 // Filter them using the function arguments
-func GetTemplates(projectStyle string, showEnabledOnly bool) ([]Template, error) {
-	req, err := http.NewRequest("GET", config.PFEOrigin()+"/api/v1/templates", nil)
+func GetTemplates(projectStyle string, showEnabledOnly bool, conID string) ([]Template, error) {
```
liamchampton: to shorten this, you could group the input vars together for the strings like this.. 
`GetTemplates(projectStyle, conID string, showEnabledOnly bool)` 
This is a bit of a fussy comment but its good practice

pkg/apiroutes/templates.go
```diff
@@ -127,7 +140,7 @@ func GetTemplateRepos() ([]utils.TemplateRepo, error) {
 
 // AddTemplateRepo adds a template repo to PFE and
 // returns the new list of existing repos
-func AddTemplateRepo(URL, description string, name string) ([]utils.TemplateRepo, error) {
+func AddTemplateRepo(URL, description string, name string, conID string) ([]utils.TemplateRepo, error) {
```
liamchampton: This can be written neater as - `AddTemplateRepo(URL, description, name, conID string)`. No need to always define the data type if they are all `string`. I also notice `URL` isn't set but the rest are, thus introducing a mixture

pkg/apiroutes/templates.go
```diff
@@ -166,17 +184,22 @@ func AddTemplateRepo(URL, description string, name string) ([]utils.TemplateRepo
 
 // DeleteTemplateRepo deletes a template repo from PFE and
 // returns the new list of existing repos
-func DeleteTemplateRepo(URL string) ([]utils.TemplateRepo, error) {
+func DeleteTemplateRepo(URL string, conID string) ([]utils.TemplateRepo, error) {
```
liamchampton: Same as previous comment on setting data types when they are all the same

pkg/actions/templates.go
```diff
@@ -24,9 +25,11 @@ import (
 // ListTemplates lists project templates of which Codewind is aware.
 // Filter them by providing flags
 func ListTemplates(c *cli.Context) {
+	conID := strings.TrimSpace(strings.ToLower(c.String("conid")))
 	templates, err := apiroutes.GetTemplates(
 		c.String("projectStyle"),
```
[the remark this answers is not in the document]
james-wallis: I've changed this function to:
```
projectStyle := c.String("projectStyle")
	conID := strings.TrimSpace(strings.ToLower(c.String("conid")))
	showEnabledOnly := c.Bool("showEnabledOnly")
	templates, templatesErr := apiroutes.GetTemplates(projectStyle, conID, showEnabledOnly)
```

pkg/actions/templates.go
```diff
@@ -82,14 +89,15 @@ func AddTemplateRepo(c *cli.Context) {
 // DeleteTemplateRepo deletes the provided template repo from PFE.
 func DeleteTemplateRepo(c *cli.Context) {
 	url := c.String("url")
-	extensions, err := apiroutes.GetExtensions()
+	conID := strings.TrimSpace(strings.ToLower(c.String("conid")))
+	extensions, err := apiroutes.GetExtensions(conID)
 	if err == nil {
-		repos, err2 := apiroutes.GetTemplateRepos()
+		repos, err2 := apiroutes.GetTemplateRepos(conID)
```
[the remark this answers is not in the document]
james-wallis: I've changed this error to `reposErr`

pkg/apiroutes/templates.go
james-wallis: done

pkg/apiroutes/templates.go
james-wallis: done

pkg/apiroutes/templates.go
james-wallis: done

pkg/actions/commands.go
```diff
@@ -291,6 +291,12 @@ func Commands() {
 							Name:  "showEnabledOnly",
 							Usage: "Filter by whether a template is enabled or not",
 						},
+						cli.StringFlag{
```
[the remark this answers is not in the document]
james-wallis: Going to leave it as a local flag as it shows up in the help better this way and its also only used in specific commands.

pkg/apiroutes/templates.go
markcor11: personally preference I guess - I dont like that shorthand myself,  if we're doing it to shorten the number of args then it might be better to wrap them in a struct and pass that instead as one single arg.

pkg/actions/templates.go
```diff
@@ -24,12 +25,12 @@ import (
...
-	templates, err := apiroutes.GetTemplates(
-		c.String("projectStyle"),
-		c.Bool("showEnabledOnly"),
-	)
-	if err != nil {
-		log.Printf("Error getting templates: %q", err)
+	projectStyle := c.String("projectStyle")
+	conID := strings.TrimSpace(strings.ToLower(c.String("conid")))
+	showEnabledOnly := c.Bool("showEnabledOnly")
+	templates, templatesErr := apiroutes.GetTemplates(projectStyle, conID, showEnabledOnly)
```
hhellyer: Sometimes the connection id is the first parameter, sometimes it's in the middle and sometimes it's that last.
Since it's required by all the functions in apiroutes it would be good if it was always in the same place - probably as the first parameter since all api calls are made on a connection.

pkg/actions/templates.go
```diff
@@ -24,12 +25,12 @@ import (
...
 // Filter them by providing flags
 func ListTemplates(c *cli.Context) {
-	templates, err := apiroutes.GetTemplates(
-		c.String("projectStyle"),
-		c.Bool("showEnabledOnly"),
-	)
-	if err != nil {
-		log.Printf("Error getting templates: %q", err)
+	projectStyle := c.String("projectStyle")
+	conID := strings.TrimSpace(strings.ToLower(c.String("conid")))
```
hhellyer: You could pull `strings.TrimSpace(strings.ToLower(c.String("conid")))` out to a utility function since it is called so often.

pkg/actions/templates.go
james-wallis: Have changed so that conid is always the first parameter

PR #223, opened 2019-11-18
Add Username to connection and remove --username flag from Bind and Sync

james-wallis:
### Summary
* Completes enhancement https://github.com/eclipse/codewind/issues/1104
* Adds a Username field to the connection struct
* Removes the need for a `--username` flag on sync, bind and in the future any other commands which use the `sechttp` library to communicate with PFE remotely.
* Adds new `PFEOriginFromConnection` function which can be used in place of `PFEOrigin` when the connection information is already in scope of a function (`connectionByID` has already been called).

#### Additional
* Fixes logging output when connection already exists when a `connection add` is attempted - output was returning a random ID rather than one for the existing connection.
* Fixes logic for previous PR when setting `https` for local PFE url.

### Testing
Manual tests to ensure bind and sync still work.
@markcor11 has also manually ran this code to see that it still works on remote deployments. 
I've also added a function for getting the PFEOrigin when the connection object already exists.

### DIFF
--- a/pkg/actions/commands.go
+++ b/pkg/actions/commands.go
@@ -88,7 +88,6 @@ func Commands() {
 						cli.StringFlag{Name: "type, t", Usage: "The type of the project", Required: true},
 						cli.StringFlag{Name: "path, p", Usage: "The path to the project", Required: true},
 						cli.StringFlag{Name: "conid", Value: "local", Usage: "The connection id for the project", Required: false},
-						cli.StringFlag{Name: "username,u", Usage: "Account Username", Required: false},
 					},
 					Action: func(c *cli.Context) error {
 						ProjectBind(c)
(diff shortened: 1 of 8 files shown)


james-wallis:
> Tried installing a remote PFE instance - worked ok, also tried the sec commands to check they still worked - they did, however on running the Connections tests this PR seems to have broken them :
> 
> ```
> # github.com/eclipse/codewind-installer/pkg/connections
> import cycle not allowed in test
> package github.com/eclipse/codewind-installer/pkg/connections (test)
>         imports github.com/eclipse/codewind-installer/pkg/apiroutes
>         imports github.com/eclipse/codewind-installer/pkg/config
>         imports github.com/eclipse/codewind-installer/pkg/connections
> FAIL    github.com/eclipse/codewind-installer/pkg/connections [setup failed]
> ```
> 
> Could you check that please

Fixed this - was a broken import after having moved the gatekeeper into its own package.

PR #242, opened 2019-11-26
{WIP}set logging stream to stdout for logr.info

liamchampton:
**Problem https://github.com/eclipse/codewind/issues/1276**
By default logrus sends output down stderr. For logr.info we need to send it to stdout.
Seen in logrus docs:
```
 // Output to stdout instead of the default stderr
  // Can be any io.Writer, see below for File example
  log.SetOutput(os.Stdout)
```

**Solution**
Set the output stream for default log level (info) to be stdout.

This will only effect `logrus` logging. No json output should be effected by this change or by `logrus`

### DIFF
--- a/pkg/actions/commands.go
+++ b/pkg/actions/commands.go
@@ -13,7 +13,6 @@ package actions
 
 import (
 	"crypto/tls"
-	"log"
 	"net/http"
 	"os"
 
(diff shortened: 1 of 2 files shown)


makandre:
If I ran:

```
cwctl --json --insecure project create /Users/andrewmak/codewind-workspace/appython01 > output.txt
```

Now my output.txt contains info messages in addition to the json output, which I think the IDEs will throw when parsing it.

tobespc:
before merging, should check with the IDE's that this is ok for them

liamchampton:
This PR fixed logrus output that has since been reverted.

PR #245, opened 2019-11-27
Revert "Add logrus for CLI output"

liamchampton:
Reverting output to remove logrus
Reverts eclipse/codewind-installer#209


tobespc:
force merging without signing as need to revert this asap

PR #261, opened 2019-12-04
cwctl version command

jcockbain:
## Summary

Adds a `cwctl` version command, which for a given connection ID returns the deployed versions of each Codewind container. Details of this are given at https://github.com/codewind-resources/design-documentation/blob/master/codewindServer/Versioning.

This PR is dependant on: https://github.com/eclipse/codewind/pull/1359

## Testing
A remote Codewind was deployed to docker desktop Kubernetes. The version command returned the versions for each container, and the time of each image build. 

Units tests will be added once we've worked on mocking DispatchHTTPRequest().

### DIFF
--- a/pkg/apiroutes/version.go
+++ b/pkg/apiroutes/version.go
@@ -1,0 +1,136 @@
+/*******************************************************************************
+ * Copyright (c) 2019 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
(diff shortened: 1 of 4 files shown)


pkg/apiroutes/version.go
```diff
@@ -0,0 +1,144 @@
...
+
+	version, err := getVersionFromEnvAPI(req, connection, HTTPClient)
+	if err != nil {
+		return "", err
+	}
+	return version, err
+}
+
+func getVersionFromEnvAPI(req *http.Request, connection *connections.Connection, HTTPClient utils.HTTPClient) (string, error) {
+	client := &http.Client{}
```
markcor11: http client is being passed in as a function parameter,  seems we dont need to create a NEW client here.   removing this line would also allow us to send in a mock client when writing tests.

jcockbain: Amazing, hopefully this is why I was having problems mocking for the tests

PR #264, opened 2019-12-04
Trim Codewind roles and update secret verbs

johnmcollier:
The roles that we assign Codewind come from Microclimate and are much more than we actually need, so this PR trims them down a bit:
- Removes any roles needed for managing CRDs or certificates right now
    - This originates from the `microclimate-devops` service account which interacted with those resources
- Updates the secret verbs to add ones necessary for Helm 3 (see https://github.com/eclipse/codewind/pull/1358)

### DIFF
--- a/pkg/remote/deploy_pfe_rbac.go
+++ b/pkg/remote/deploy_pfe_rbac.go
@@ -37,7 +37,7 @@ func CreateCodewindRoles(deployOptions *DeployOptions) rbacv1.ClusterRole {
 		rbacv1.PolicyRule{
 			APIGroups: []string{""},
 			Resources: []string{"secrets"},
-			Verbs:     []string{"get", "list", "create", "watch", "delete"},
+			Verbs:     []string{"get", "list", "create", "watch", "delete", "delete", "patch", "update"},
 		},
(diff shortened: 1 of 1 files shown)


pkg/remote/deploy_pfe_rbac.go
```diff
@@ -37,7 +37,7 @@ func CreateCodewindRoles(deployOptions *DeployOptions) rbacv1.ClusterRole {
 		rbacv1.PolicyRule{
 			APIGroups: []string{""},
 			Resources: []string{"secrets"},
-			Verbs:     []string{"get", "list", "create", "watch", "delete"},
+			Verbs:     []string{"get", "list", "create", "watch", "delete", "delete", "patch", "update"},
```
elsony: Is delete duplicated?

johnmcollier: Good catch, that was an oversight on my part. Fixed!

maysunfaisal:
meant to say, looks good as per https://github.com/eclipse/codewind-che-plugin/pull/123

Wrong link in approval msg

PR #265, opened 2019-12-04
Allow Codewind to deploy on Kubes with auth plugins

johnmcollier:
On certain Kubes that use auth plugins for Kubernetes (such as IKS, GKE, etc), `cwctl install remote ...` will fail with the following errors:
```
INFO[0000] Unable to retrieve Kubernetes clientset No Auth Provider found for name "oidc"
ERRO[0000] Error: rem_not_found - No Auth Provider found for name "oidc"
``` 
or on GKE
```
INFO[0000] Unable to retrieve Kubernetes clientset No Auth Provider found for name "gcp"
ERRO[0000] Error: rem_not_found - No Auth Provider found for name "gcp"
```

The solution here is to add the plugins from client-go as blank imports to whatever module is creating the Kube client (in this case deploy.go) (see https://github.com/kubernetes/client-go/issues/242#issuecomment-478423447 for more details).

Verified on IKS and my personal GKE cluster

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,6 +2,14 @@
 
 
 [[projects]]
+  digest = "1:26ee1e365ea8f312ee11e170fc6675bac0dd3d4adf2406e753d0a43527e1afb8"
+  name = "cloud.google.com/go"
+  packages = ["compute/metadata"]
+  pruneopts = "UT"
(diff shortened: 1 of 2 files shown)


johnmcollier:
@markcor11 Please review, I also updated the Gopkg.lock, as new packages were pulled in

PR #277, opened 2019-12-09
Allow deployments to be created as unprivileged

johnmcollier:
For https://github.com/eclipse/codewind/issues/1437

As part of a holdover from the code that's used to deploy Codewind on Che, `cwctl` remote deploy code is creating every deployment as privileged, when only PFE needs to be privileged.

This PR adds a a new boolean parameter to `generateDeployment` called `privileged`, when set to true it creates a privileged deployment, otherwise it creates an unprivileged one

**Keycloak**
```
johns-mbp-3:codewind-installer johncollier$ printf "$(k get po codewind-keycloak-k3yr3bx4-8fd96469-4tstd -o jsonpath="{.spec.containers[0].securityContext.privileged}")\n"
false
```

**PFE**
```
johns-mbp-3:codewind-installer johncollier$ printf "$(k get po codewind-pfe-k3yr3bx4-6974c78b79-lbgms -o jsonpath="{.spec.containers[0].securityContext.privileged}")\n"
true
```


johnmcollier:
@markcor11 PTAL

PR #283, opened 2019-12-10
Delete by container name not image

james-wallis:
**Fixes https://github.com/eclipse/codewind/issues/1388**
### Summary
* Removes by container name rather than image name.
  * Appsody image names are not build by us and don't contain the `cw-` prefix.
* Adds function to get containers to remove and a test to check its functionality
* Add function to check a containers `AutoRemove` status so we only remove ones that do not `AutoRemove` themselves. 

### Testing
* Added test for getting containes to delete.
* Manually tested that `cwctl` deletes Appsody containers.

### DIFF
--- a/pkg/actions/stop-all.go
+++ b/pkg/actions/stop-all.go
@@ -15,31 +15,19 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/docker/docker/api/types"
 	"github.com/eclipse/codewind-installer/pkg/utils"
 )
 
 //StopAllCommand to stop codewind and project containers
 func StopAllCommand() {
-	containerArr := []string{
-		"codewind-pfe",
-		"codewind-performance",
-		"cw-",
-		"appsody",
-	}
-
 	containers := utils.GetContainerList()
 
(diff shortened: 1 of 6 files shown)


pkg/actions/stop-all.go
```diff
@@ -52,3 +40,24 @@ func StopAllCommand() {
...
 }
+
+func getContainersToRemove(containerList []types.Container) []types.Container {
+	codewindContainerNames := []string{
+		"codewind-pfe",
+		"codewind-performance",
+	}
+
+	// Docker returns all the names with a "/" on the front
+	projectContainerPrefix := "/cw-"
```
makandre: Doesn't that mean it should be `/codewind-pfe` and `/codewind-performance` above then?

james-wallis: It doesn't in the current design but looking at the container names we use (codewind-pfe-amd64, codewind-performance-amd64) they can also use prefix rather than contains too.
I'll make that change

james-wallis: @makandre See latest change.

PR #294, opened 2019-12-13
Use docker-compose to bring down containers/remove images

liamchampton:
**Problem https://github.com/eclipse/codewind/issues/1148**
We currently mix technologies. We use `docker-compose up` to bring up containers but we then use docker API's to shut down the containers and remove the images. This leaves cached data from the `docker-compose up` and seems to be causing the issue 
```
ERROR: for codewind-performance  b'no such image: sha256:f0a....
```

**Solution in this PR**
1) Remove the re-tagging of images so the image name is consistently `eclipse/codewind-...` throughout. This will also mean that if `start` fails because of a wrong tag being specified, it will attempt to pull down the image from docker hub if it doesnt already exist on the host machine. Will fix this issue https://github.com/eclipse/codewind/issues/462
2) Rename `codewind-docker-compose.yaml` -> `docker-compose.yaml`
3) On first start of Codewind, write the `docker-compose.yaml` to the `./codewind` directory and keep it stored in here. No longer delete it so it can be used by other `docker-compose` commands
4) Use `docker-compose -f <compose-file> rm --stop -f` to stop only the `codewind-pfe` & `codewind-performance` containers. Project containers will continue to be stopped using a docker API since they are not specified in the `docker-compose.yaml` services
5) Use `docker-compose -f <compose-file> down --rmi all` to remove `eclipse/codewind-pfe` & `eclipse/codewind-performance` images. Project images will continue to be removed using the docker API as they are not specified in the `docker-compose.yaml` services
6) Since this will use `docker-compose` to remove the images, the `remove` command *should* have a tag specified. If no tag is provided it will default to `latest`

**Tested**
*Operating Systems*
1) MacOS - `install/start/stop/stop-all/remove` working
2) Windows - `install/start/stop/stop-all/remove` working
3) Linux - `install/start/stop/stop-all/remove` working

*Bats test output*
`18 tests, 0 failures, 6 skipped`

**Extra**
- Better error checking and handling will be provided in a follow up PR from @jcockbain 
- The logic of the stopping of the containers in this PR also fixes https://github.com/eclipse/codewind/issues/1502

### DIFF
--- a/pkg/actions/commands.go
+++ b/pkg/actions/commands.go
@@ -15,27 +15,15 @@ import (
 	"crypto/tls"
 	"net/http"
 	"os"
-	"runtime"
 
 	"github.com/eclipse/codewind-installer/pkg/appconstants"
+	desktoputils "github.com/eclipse/codewind-installer/pkg/desktop_utils"
 	"github.com/eclipse/codewind-installer/pkg/errors"
 	logr "github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 )
 
-// Get home directory
-func getHomeDir() string {
-	homeDir := ""
-	const GOOS string = runtime.GOOS
-	if GOOS == "windows" {
-		homeDir = os.Getenv("USERPROFILE")
-	} else {
-		homeDir = os.Getenv("HOME")
-	}
-	return homeDir
-}
-
-var homeDir = getHomeDir()
+var homeDir = desktoputils.GetHomeDir()
 var dockerComposeFile = homeDir + "/.codewind/docker-compose.yaml"
 
 const healthEndpoint = "/api/v1/environment"
--- a/pkg/actions/commands.go
+++ b/pkg/actions/commands.go
@@ -15,6 +15,7 @@ import (
 	"crypto/tls"
 	"net/http"
 	"os"
+	"runtime"
 
 	"github.com/eclipse/codewind-installer/pkg/appconstants"
 	"github.com/eclipse/codewind-installer/pkg/errors"
@@ -22,7 +23,20 @@ import (
 	"github.com/urfave/cli"
 )
 
-var tempFilePath = "codewind-docker-compose.yaml"
+// Get home directory
+func getHomeDir() string {
+	homeDir := ""
+	const GOOS string = runtime.GOOS
+	if GOOS == "windows" {
+		homeDir = os.Getenv("USERPROFILE")
+	} else {
+		homeDir = os.Getenv("HOME")
+	}
+	return homeDir
+}
+
+var homeDir = getHomeDir()
+var dockerComposeFile = homeDir + "/.codewind/docker-compose.yaml"
 
 const healthEndpoint = "/api/v1/environment"
 
@@ -228,7 +242,7 @@ func Commands() {
 				},
 			},
 			Action: func(c *cli.Context) error {
-				StartCommand(c, tempFilePath, healthEndpoint)
+				StartCommand(c, dockerComposeFile, healthEndpoint)
 				return nil
 			},
 		},
@@ -256,7 +270,7 @@ func Commands() {
 			Name:  "stop",
 			Usage: "Stop the running Codewind containers",
 			Action: func(c *cli.Context) error {
-				StopCommand()
+				StopCommand(c, dockerComposeFile)
 				return nil
 			},
 		},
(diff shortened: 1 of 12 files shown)


pkg/actions/commands.go
```diff
@@ -15,14 +15,28 @@ import (
...
 
 	"github.com/eclipse/codewind-installer/pkg/appconstants"
 	"github.com/eclipse/codewind-installer/pkg/errors"
 	logr "github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 )
 
-var tempFilePath = "codewind-docker-compose.yaml"
+// Get home directory
+func getHomeDir() string {
```
hhellyer: Pulling this out to a function is good but there's about 4 other places we get the homeDir now. Could we make it a global function/variable to make sure we do this the same way everywhere?

hhellyer: Update, alternatively if/when we are on Go 12 could we just use: https://golang.org/pkg/os/#UserHomeDir ?

pkg/actions/install.go
```diff
@@ -32,11 +32,10 @@ func InstallCommand(c *cli.Context) {
...
 
-	imageArr := [2]string{"docker.io/eclipse/codewind-pfe-amd64:",
-		"docker.io/eclipse/codewind-performance-amd64:"}
-
-	targetArr := [2]string{"codewind-pfe-amd64:",
-		"codewind-performance-amd64:"}
+	imageArr := [2]string{
+		"docker.io/eclipse/codewind-pfe-amd64:",
+		"docker.io/eclipse/codewind-performance-amd64:",
+	}
```
hhellyer: Could we define this as:
```
	imageArr := [2]string{
		"docker.io/eclipse/codewind-pfe-amd64:" + tag,
		"docker.io/eclipse/codewind-performance-amd64:" + tag,
	}
```
I don't think we can use the names without the tag ever as they end in `:` so we can get away with just appending tag once and reduce the scope for forgetting.

pkg/utils/docker.go
```diff
@@ -106,64 +106,106 @@ const (
...
+	cmd := exec.Command("docker-compose", "-f", dockerComposeFile, "up", "-d", "--force-recreate")
+	output := new(bytes.Buffer)
+	cmd.Stdout = output
+	cmd.Stderr = output
+	if err := cmd.Start(); err != nil { // after 'Start' the program is continued and script is executing in background
+		DeleteTempFile(dockerComposeFile)
+		errors.CheckErr(err, 101, "Is docker-compose installed?")
+	}
+	fmt.Printf("Please wait whilst containers initialize... %s \n", output.String())
+	cmd.Wait()
```
hhellyer: Should we be checking the return code from Wait() to get the exit code for the process? Wait() will return an error if the command failed.
We check for certain known error strings but we don't check the commands exit code.

pkg/utils/docker.go
```diff
@@ -106,64 +106,106 @@ const (
...
+	setupDockerComposeEnvs(tag, "stop")
+	cmd := exec.Command("docker-compose", "-f", dockerComposeFile, "rm", "--stop", "-f")
+	output := new(bytes.Buffer)
+	cmd.Stdout = output
+	cmd.Stderr = output
+	if err := cmd.Start(); err != nil { // after 'Start' the program is continued and script is executing in background
+		errors.CheckErr(err, 101, "")
+	}
+	fmt.Printf("Please wait whilst containers shutdown... %s \n", output.String())
+	cmd.Wait()
```
hhellyer: See above comment about wait.

pkg/utils/docker.go
```diff
@@ -39,7 +39,7 @@ var data = `
 version: 2
 services:
  codewind-pfe:
-  image: ${REPOSITORY}codewind-pfe${PLATFORM}:${TAG}
+  image: eclipse/codewind-pfe${PLATFORM}:${TAG}
```
hhellyer: Since we use these image names several times in this file could we define constants for the names?
https://github.com/eclipse/codewind-installer/pull/294/files#diff-02a4c9edab65df5dfdd065f5b995d1faR510
This could then be `${PFE_IMAGE_NAME}${PLATFORM}:${TAG}`
And the array of names could be:
```
baseImageNameArr = []string{
  PFE_IMAGE_NAME,
  PERFORMANCE_IMAGE_NAME,
}
```

pkg/actions/commands.go
liamchampton: Sadly not on go 12 (YET!), hence why we are doing this, but i agree. Will move this into a global func

PR #302, opened 2019-12-18
Fix Linter warnings

markcor11:
## Enhancement

Resolves various Linter warnings

Signed-off-by: markcor11 <mark.cornaia@uk.ibm.com>

### DIFF
--- a/pkg/apiroutes/templates.go
+++ b/pkg/apiroutes/templates.go
@@ -229,7 +229,7 @@ func DeleteTemplateRepo(conID, URL string) ([]utils.TemplateRepo, error) {
 		return nil, conErr.Err
 	}
 
-	req, err := http.NewRequest("DELETE", conURL+"/api/v1/templates/repositories", bytes.NewBuffer(jsonValue))
+	req, _ := http.NewRequest("DELETE", conURL+"/api/v1/templates/repositories", bytes.NewBuffer(jsonValue))
 	req.Header.Set("Content-Type", "application/json")
 	client := &http.Client{}
 	resp, httpSecError := sechttp.DispatchHTTPRequest(client, req, conInfo)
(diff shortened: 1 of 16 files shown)


pkg/apiroutes/templates.go
```diff
@@ -229,7 +229,7 @@ func DeleteTemplateRepo(conID, URL string) ([]utils.TemplateRepo, error) {
 		return nil, conErr.Err
 	}
 
-	req, err := http.NewRequest("DELETE", conURL+"/api/v1/templates/repositories", bytes.NewBuffer(jsonValue))
+	req, _ := http.NewRequest("DELETE", conURL+"/api/v1/templates/repositories", bytes.NewBuffer(jsonValue))
```
liamchampton: if `err` is ignored and there is an issue with this request it could panic. 
A simple local test by trying a bad request and ignoring the error locally returns
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x122f048]
```
I think the  `warning: ineffectual assignment to err (ineffassign)` is just because `err` isnt being checked. Is it worth checking this before doing the `sechttp.DispatchHTTPRequest`?

pkg/apiroutes/templates.go
```diff
@@ -331,7 +331,7 @@ func BatchPatchTemplateRepos(conID string, operations []RepoOperation) ([]SubRes
 		return nil, conErr.Err
 	}
 
-	req, err := http.NewRequest("PATCH", conURL+"/api/v1/batch/templates/repositories", bytes.NewBuffer(jsonValue))
+	req, _ := http.NewRequest("PATCH", conURL+"/api/v1/batch/templates/repositories", bytes.NewBuffer(jsonValue))
```
liamchampton: Same as above

pkg/apiroutes/templates.go
markcor11: Thats correct, the CLI will crash in exactly the same way it crashes before. Changing the logic flow to report back an error (which will require a change to the function signature if we are to use our standard package errors, and handling of the error) is beyond the scope of this PR.  I didn't want to modify the logic but rather call out that err wasn't being processed.  I think it's easier to scan for "_" than for "err" when looking for holes.

pkg/apiroutes/templates.go
markcor11: Same as above

PR #303, opened 2019-12-18
Use DockerError struct for all docker errors

jcockbain:
### Summary

Updates `docker.go` functions to return a `DockerError` struct, and process this structure in the actions calling them.

Outputs json errors, when the`--json` global flag is set. 

Uses logrus for non-json errors, as opposed to fmt. 

Current output:

Screenshot 2019-12-18 at 11 13 42

New Output (--json not set):

Screenshot 2019-12-18 at 11 09 18

New Output (--json set):

Screenshot 2019-12-18 at 11 09 37

### Testing

All existing tests pass locally. 

All commands that call the modified actions run successfully, and output the expected errors (shown above) when my docker is turned off.

### DIFF
--- a/pkg/actions/start.go
+++ b/pkg/actions/start.go
@@ -16,20 +16,14 @@ import (
 	"os"
 
 	"github.com/eclipse/codewind-installer/pkg/utils"
-	logr "github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 )
 
 // StartCommand : start the codewind containers
 func StartCommand(c *cli.Context, dockerComposeFile string, healthEndpoint string) {
-	printAsJSON := c.GlobalBool("json")
 	status, err := utils.CheckContainerStatus()
 	if err != nil {
-		if printAsJSON {
-			fmt.Println(err.Error())
-		} else {
-			logr.Error(err.Desc)
-		}
+		HandleDockerError(err)
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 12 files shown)


pkg/utils/docker_error.go
```diff
@@ -21,9 +21,24 @@ type DockerError struct {
 }
 
 const (
-	errOpValidate         = "docker_validate" // validate docker images
-	errOpClientCreate     = "CLIENT_CREATE_ERROR"
-	errOpContainerInspect = "CONTAINER_INSPECT_ERROR"
+	errOpValidate            = "DOCKER_VALIDATE"
+	errOpClientCreate        = "CLIENT_CREATE_ERROR"
+	errOpContainerInspect    = "CONTAINER_INSPECT_ERROR"
+	errOpContainersStopped   = "CONTAINERS_STOPPED"
```
liamchampton: is there a reason this is plural (the rest are singular) and doesn't include the suffix `ERROR`? This and `errOpStopContainer       = "CONTAINER_STOP_ERROR"` are very similar, could these be put into one?

jcockbain: No reason, good spot

pkg/utils/utils_test.go
```diff
@@ -40,14 +40,20 @@ func TestRemoveImage(t *testing.T) {
...
+		t.Fail()
+	}
 	assert.Equal(t, result, false, "should return false: no images are installed")
 }
 
 func TestCheckContainerStatusFalse(t *testing.T) {
 	// Test checks that container list can be searched
 	// False return as no containers have been started for this test
-	result := CheckContainerStatus()
+	result, err := CheckContainerStatus()
```
hhellyer: The test could type the `err` variable to make sure a docker error is returned.

pkg/actions/start.go
```diff
@@ -13,14 +13,25 @@ package actions
...
-	status := utils.CheckContainerStatus()
+	printAsJSON := c.GlobalBool("json")
+	status, err := utils.CheckContainerStatus()
+	if err != nil {
+		if printAsJSON {
+			fmt.Println(err.Error())
+		} else {
+			logr.Error(err.Desc)
+		}
+		os.Exit(1)
```
hhellyer: This is done so frequently it would be nice to see it pulled out to a utility function.
Something like:
```suggestion
if err != nil {
		printAsJSON := c.GlobalBool("json")
		if printAsJSON {
			fmt.Println(err.Error())
		} else {
			logr.Error(err.Desc)
		}
		os.Exit(1)
}
```
There will still be a few places that don't use this but there are also follow up PR's like: https://github.com/eclipse/codewind-installer/pull/316
that would benefit from this.

PR #315, opened 2020-01-07
Delete inf file after upgrade

stalleyj:
Signed-off-by: Julie Stalley <julie_stalley@uk.ibm.com>

Add delete of the inf file after we have successfully migrated a project


stalleyj:
We don't care if we can't delete the inf file so no point catching/reporting the error

PR #317, opened 2020-01-07
make project create consistent

jcockbain:
Resolves: https://github.com/eclipse/codewind/issues/1452

## Summary

Makes the `cwctl project create` command consistent with the other commands. 

- Splits out create and validate into their own subcommands, at the moment this is decided based on whether a URL is given.

- Uses `--path`, as opposed to the first non-flagged argument for path. 

- Adds documentation for the command in the README.md and commands file.

## Testing

- Commands run as expected. 

- Tests pass locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,38 +183,33 @@ Subcommands:</br>
 > --path,-p value Project Path
 
 `bind` - Bind a project to Codewind for building and running
-
 > **Flags:**
-> --name,-n value Project name
-> --language,-l value Project language
-> --type,-t value Project Type
-> --path,-p value Project Path
-> --conid value Connection ID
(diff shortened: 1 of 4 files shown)


hhellyer:
Does this change require corresponding changes to the VSCode and Eclipse plugins? Have you tested a local build of these changes with the VSCode or Eclipse plugin?

jcockbain:
Yes, I was just making them now. Will mark as WIP until that's done

jcockbain:
Proposed changes to plugins are at https://github.com/eclipse/codewind-vscode/pull/383 and https://github.com/eclipse/codewind-eclipse/pull/516.

PR #324, opened 2020-01-10
Change cwctl remove to use sechhttp

james-wallis:
Fix for https://github.com/eclipse/codewind/issues/1660
### Summary
* Use `sechhttp` in `unbind.go` so that we can reach remote deployments.

### Testing
* Successfully ran `cwctl remove` on an OpenShift cluster (Mongrel).

### DIFF
--- a/pkg/project/getproject.go
+++ b/pkg/project/getproject.go
@@ -37,7 +37,7 @@ func GetProject(httpClient utils.HTTPClient, conID, projectID string) (*Project,
 	}
 
 	// send request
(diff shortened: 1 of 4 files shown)


james-wallis:
> This look good. Good use of the `DispatchHTTPRequest` functions. I assume all tests are still passing locally (BATS/Units)?

All passing apart from one that was broken in a separate PR, test fixed in :https://github.com/eclipse/codewind-installer/pull/329

PR #329, opened 2020-01-13
Only use global --json flag, remove print for image check

james-wallis:
### Summary
* Change `install` and `status` command to use the global `--json` flag rather than a scoped one. Previously they accepted both to allow the IDE's time to update.
    * `status` had its scoped `--json` flag removed in: https://github.com/eclipse/codewind-installer/pull/303 this PR just updates the tests for it.
* Change a `log` in `Docker.go` to be a `logr.Debug` so that it isn't outputted to the terminal - fixes it being outputted when the `--json` flag is given.
   * Previously outputted: 
```
❯ cwctl --json status
195-90-nip-io:7443/my-rhel-icp-admin/jamesw0001
2020/01/13 11:16:04 No tag available. Defaulting to ''
2020/01/13 11:16:04 No tag available. Defaulting to ''
{"status":"stopped","installed-versions":[]}
```

### Testing
Ran (and fixed) that `bats` tests.

### Note
I have checked the Eclipse and VSCode plugins and they should use the global `--json` option but @eharris369 and @tetchel could you confirm that this PR won't break the status or install commands?

### DIFF
--- a/integration.bats
+++ b/integration.bats
@@ -1,16 +1,16 @@
 #!/usr/bin/env bats
 
-@test "invoke install command - install latest with --json" {
+@test "invoke install command - install latest with --json global flag set" {
   cd cmd/cli/
-  run go run main.go install --json
+  run go run main.go --json install
   echo "status = ${status}"
   echo "output trace = ${output}"
     [ "$status" -eq 0 ]
 }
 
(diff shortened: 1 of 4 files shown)


eharris369:
This will not break Eclipse.  I fixed the code to pass the json option as a global option a while ago: https://github.com/eclipse/codewind-eclipse/pull/499.

PR #335, opened 2020-01-16
Pass the logging level to the PFE container.

hhellyer:
This PR enables logging from startup by setting the log level environment variable when the PFE docker container is started. This is needed to get full logging information for initialisation from PFE as otherwise the log level can only be adjusted once the container has started and the REST API is available.

As there was already a `loglevel` flag for `cwctl` that took the same log levels this change just passes through that same level to PFE rather than have multiple flags. (If we are getting someone to trace startup it probably makes sense to get the logging for `cwctl` and the PFE container as they are closely integrated.)


hhellyer:
This was for https://github.com/eclipse/codewind/issues/729

PR #341, opened 2020-01-21
Prevent CheckContainerStatus() picking up images not started by docker-compose.

hhellyer:
# Description of pull request

We are unable to run codewind locally in docker and docker's local kubernetes cluster. The containers do not clash but we check if codewind is running by looking at the image names, which are being used in the docker containers created by kubernetes not the container names which do not match what we use in our docker-compose.yaml

## Solution

This updates the `CheckContainerStatus()` function to look at the container names, which are unique to local codewind, not the image names which are shared.
Previously it was possible to run both versions at the same time, provided you started codewind in kube after you'd started it in docker. You then couldn't restart codewind in docker as it would find the containers running our images in local kubernetes which was a problem for development.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Testing undertaken
I've started and stopped codewind locally while also running codewind installed into dockers local kubernetes cluster.
I've connected to both at the same time.

## Checklist

- [X] I have commented my code where needed
- [ ] I have made corresponding changes to the documentation
- [X] My changes do not generate any new warnings/linter errors
- [] If necessary, I have added tests that prove my fix is effective
- [X] New and existing unit tests pass locally with my changes
- [ ] There are no typos in the code comments or this pull request
- [X] I have signed my commits and conformed with the Eclipse commit record guidelines

## Extra

Eclipse commit record guidelines followed <https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#The_Commit_Record>

### DIFF
--- a/pkg/utils/docker.go
+++ b/pkg/utils/docker.go
@@ -44,7 +44,7 @@ var baseImageNameArr = [2]string{
 const pfeContainerName = "codewind-pfe"
 const performanceContainerName = "codewind-performance"
 
-var containerImageNames = []string{
+var containerImageNames = [...]string{
 	pfeContainerName,
 	performanceContainerName,
 }
--- a/pkg/utils/docker.go
+++ b/pkg/utils/docker.go
@@ -41,11 +41,19 @@ var baseImageNameArr = [2]string{
 	performanceImageName,
 }
 
+const pfeContainerName = "codewind-pfe"
+const performanceContainerName = "codewind-performance"
+
+var containerImageNames = []string{
+	pfeContainerName,
+	performanceContainerName,
+}
+
 // codewind-docker-compose.yaml data
 var data = `
 version: 2
 services:
- codewind-pfe:
+ ` + pfeContainerName + `:
   image: ${PFE_IMAGE_NAME}${PLATFORM}:${TAG}
   container_name: codewind-pfe
   user: root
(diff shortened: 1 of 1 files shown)


pkg/utils/docker.go
```diff
@@ -41,11 +41,19 @@ var baseImageNameArr = [2]string{
...
+	pfeContainerName,
+	performanceContainerName,
+}
+
 // codewind-docker-compose.yaml data
 var data = `
 version: 2
 services:
- codewind-pfe:
+ ` + pfeContainerName + `:
```
liamchampton: Why have you added spaces around the name of the service?

pkg/utils/docker.go
```diff
@@ -41,11 +41,19 @@ var baseImageNameArr = [2]string{
 	performanceImageName,
 }
 
+const pfeContainerName = "codewind-pfe"
+const performanceContainerName = "codewind-performance"
+
+var containerImageNames = []string{
```
liamchampton: Since we know the number of indexes in this array it is good practice to specify the length. Should be written as `[2]string{foo,bar}`

hhellyer: I've re-written it as `[...]` so the compiler calculates the length.
I didn't realise using [] actually declared a slice and this list of names doesn't need modifying.

PR #356, opened 2020-01-31
Add IKS annotations for HTTPs services

markcor11:
## Problem 

Codewind Gatekeeper is not accessible in IKS but works ok on Openshift in IBM Cloud. 

This happens because by default, Ingresses in IKS are expecting the destination service (Gatekeeper) to be using HTTP.   Gatekeeper in IKS uses HTTPs and is thus not accessible returning a 502 gateway error as reported in issue #1900

## Solution

The Ingresses created by CWCTL are missing two required annotations for IBM Cloud.  This PR adds the two missing annotations. 

Signed-off-by: markcor11 <mark.cornaia@uk.ibm.com>

### DIFF
--- a/pkg/remote/deploy_gatekeeper.go
+++ b/pkg/remote/deploy_gatekeeper.go
@@ -168,12 +168,12 @@ func generateIngressGatekeeper(codewind Codewind) extensionsv1.Ingress {
 
 	annotations := map[string]string{
 		"nginx.ingress.kubernetes.io/rewrite-target":     "/",
+		"ingress.bluemix.net/redirect-to-https":          "True",
+		"ingress.bluemix.net/ssl-services":               "ssl-service=" + GatekeeperPrefix + "-" + codewind.WorkspaceID,
 		"nginx.ingress.kubernetes.io/backend-protocol":   "HTTPS",
 		"kubernetes.io/ingress.class":                    "nginx",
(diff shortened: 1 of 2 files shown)


johnmcollier:
@markcor11 is this PR still going in?

pkg/remote/deploy_gatekeeper.go
```diff
@@ -168,12 +168,12 @@ func generateIngressGatekeeper(codewind Codewind) extensionsv1.Ingress {
 
 	annotations := map[string]string{
 		"nginx.ingress.kubernetes.io/rewrite-target":     "/",
+		"ingress.bluemix.net/redirect-to-https":          "True",
```
johnmcollier: 1. Does adding this in by default affect installs on non-IKS / non-IBM Cloud platforms in any way?
2. We've got the ability to check if: we're running on OpenShift **and** if we're running on IBM Cloud (See function that checks for IBM File storage class). Can we set these annotations only if we're not on OpenShift **but** we're on IBM Cloud?

markcor11: Does not have any effect where that annotation is not available. I tried the changes on IBM Public Cloud + ROKS,  and elsewhere on Openshift 3.11 and Kube in Docker desktop.

PR #357, opened 2020-01-31
mock PFE calls in TestWriteNewCwSettings + add extra project unit tests 

jcockbain:

# Description of pull request

Mocks PFE when testing writeNewCwSettings, which allows the tests to run without a PFE running. 

Slight refactor of the local function to allow for the PFE client to be passed in.

This is the last piece of work to allow the project package tests to run in Jenkins.

Also adds new unit tests for functions in project/create.go.

Towards https://github.com/eclipse/codewind.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Testing undertaken

All new and old tests pass. 

Create and validate commands run successfully.

## Checklist

- [x] I have commented my code where needed
- [ ] I have made corresponding changes to the documentation
- [x] My changes do not generate any new warnings/linter errors
- [x] If necessary, I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes
- [x] There are no typos in the code comments or this pull request
- [x] I have signed my commits and conformed with the Eclipse commit record guidelines

## Extra

Eclipse commit record guidelines followed <https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#The_Commit_Record>

### DIFF
--- a/pkg/project/create.go
+++ b/pkg/project/create.go
@@ -323,9 +323,9 @@ func renameLegacySettings(pathToLegacySettings string, pathToCwSettings string) 
 
 // writeNewCwSettings writes a default .cw-settings file to the given path,
 // dependant on the build type of the project
-func writeNewCwSettings(client utils.HTTPClient, connection *connections.Connection, conURL string, pathToCwSettings string, BuildType string) *ProjectError {
+func writeNewCwSettings(httpClient utils.HTTPClient, connection *connections.Connection, conURL string, pathToCwSettings string, BuildType string) *ProjectError {
 
-	defaultCwSettings, projErr := getDefaultCwSettings(client, connection, conURL, BuildType)
+	defaultCwSettings, projErr := getDefaultCwSettings(httpClient, connection, conURL, BuildType)
 	if projErr != nil {
 		return projErr
 	}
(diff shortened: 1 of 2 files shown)


pkg/project/create.go
```diff
@@ -313,39 +323,35 @@ func renameLegacySettings(pathToLegacySettings string, pathToCwSettings string)
 
 // writeNewCwSettings writes a default .cw-settings file to the given path,
 // dependant on the build type of the project
-func writeNewCwSettings(conID string, pathToCwSettings string, BuildType string) *ProjectError {
-	defaultCwSettings, projErr := getDefaultCwSettings(conID, BuildType)
+func writeNewCwSettings(client utils.HTTPClient, connection *connections.Connection, conURL string, pathToCwSettings string, BuildType string) *ProjectError {
```
james-wallis: nit: Think `client` is `httpClient` in every other function that takes it as a parameter. also  `getDefaultCwSettings`

pkg/project/create.go
```diff
@@ -191,13 +191,23 @@ func writeCwSettingsIfNotInProject(conID string, projectPath string, BuildType s
...
+	connection, conErr := connections.GetConnectionByID(conID)
+	if conErr != nil {
+		return &ProjectError{errOpCreateProject, conErr, conErr.Desc}
+	}
+
+	conURL, configErr := config.PFEOriginFromConnection(connection)
+	if configErr != nil {
+		return &ProjectError{errOpCreateProject, conErr, configErr.Desc}
+	}
+
```
james-wallis: obv wasn't in this pr but might be better to return a different error code than `errOpCreateProject` as it doesn't say what went wrong.

Also `errOpCreateProject`'s text has a space not an underscore (if you make this change).

james-wallis: `errOpConNotFound`

james-wallis: Spoke to @jcockbain and we'll keep the operation code the same

pkg/project/create.go
jcockbain: Good find, changed

PR #390, opened 2020-02-28
Retry template request on lock

james-wallis:

## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?
Accompanies https://github.com/eclipse/codewind/pull/2346.
This PR allows the CLI to retry a HTTP request to the Templates API up to five times (const value).
This is to reduce the probability of the IDEs seeing the `423 locked` response code.

No.

### DIFF
--- a/pkg/apiroutes/templates.go
+++ b/pkg/apiroutes/templates.go
@@ -80,7 +80,7 @@ func GetTemplates(conID, projectStyle string, showEnabledOnly bool) ([]Template,
 	req.URL.RawQuery = query.Encode()
 	client := &http.Client{}
 
-	resp, httpSecError := httpRequestWithRetryOnLock(client, req, conInfo)
+	resp, httpSecError := HTTPRequestWithRetryOnLock(client, req, conInfo)
 	if httpSecError != nil {
 		return nil, httpSecError
 	}
@@ -112,7 +112,7 @@ func GetTemplateStyles(conID string) ([]string, error) {
 		return nil, err
 	}
 	client := &http.Client{}
-	resp, httpSecError := httpRequestWithRetryOnLock(client, req, conInfo)
+	resp, httpSecError := HTTPRequestWithRetryOnLock(client, req, conInfo)
 	if httpSecError != nil {
 		return nil, httpSecError
 	}
@@ -144,7 +144,7 @@ func GetTemplateRepos(conID string) ([]utils.TemplateRepo, error) {
 		return nil, err
 	}
 	client := &http.Client{}
-	resp, httpSecError := httpRequestWithRetryOnLock(client, req, conInfo)
+	resp, httpSecError := HTTPRequestWithRetryOnLock(client, req, conInfo)
 	if httpSecError != nil {
 		return nil, httpSecError
 	}
@@ -191,7 +191,7 @@ func AddTemplateRepo(conID, URL, description, name string) ([]utils.TemplateRepo
 		return nil, err
 	}
 	client := &http.Client{}
-	resp, httpSecError := httpRequestWithRetryOnLock(client, req, conInfo)
+	resp, httpSecError := HTTPRequestWithRetryOnLock(client, req, conInfo)
 	if httpSecError != nil {
 		return nil, httpSecError
 	}
@@ -234,7 +234,7 @@ func DeleteTemplateRepo(conID, URL string) ([]utils.TemplateRepo, error) {
 	req, _ := http.NewRequest("DELETE", conURL+"/api/v1/templates/repositories", bytes.NewBuffer(jsonValue))
 	req.Header.Set("Content-Type", "application/json")
 	client := &http.Client{}
-	resp, httpSecError := httpRequestWithRetryOnLock(client, req, conInfo)
+	resp, httpSecError := HTTPRequestWithRetryOnLock(client, req, conInfo)
 	if httpSecError != nil {
 		return nil, httpSecError
 	}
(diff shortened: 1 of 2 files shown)


pkg/apiroutes/templates.go
```diff
@@ -355,3 +357,22 @@ func BatchPatchTemplateRepos(conID string, operations []RepoOperation) ([]SubRes
 
 	return subResponsesFromBatchOperation, nil
 }
+
+// httpRequestWithRetryOnLock : Retries a request until either the maxHTTPRetries limit has been hit or the response status code is not equal to 423 (templates locked HTTP code)
+func httpRequestWithRetryOnLock(httpClient utils.HTTPClient, originalRequest *http.Request, connection *connections.Connection) (*http.Response, *sechttp.HTTPSecError) {
+	var maxHTTPRetries int = 5
```
jcockbain: don't need "int"

pkg/apiroutes/templates.go
```diff
@@ -355,3 +357,22 @@ func BatchPatchTemplateRepos(conID string, operations []RepoOperation) ([]SubRes
 
 	return subResponsesFromBatchOperation, nil
 }
+
+// httpRequestWithRetryOnLock : Retries a request until either the maxHTTPRetries limit has been hit or the response status code is not equal to 423 (templates locked HTTP code)
+func httpRequestWithRetryOnLock(httpClient utils.HTTPClient, originalRequest *http.Request, connection *connections.Connection) (*http.Response, *sechttp.HTTPSecError) {
```
jcockbain: if you did want to look at tests for this, an interface with a Calls counter could come in use? See halfway down this page https://quii.gitbook.io/learn-go-with-tests/go-fundamentals/mocking. Could be overkill, but then could come in useful in other areas too.

jcockbain: You could then modify the response on dispatchRequest based on the number of calls

james-wallis: This is a good suggestion but I think at the moment this is overkill. I have added tests which ensure that the correct response and errors are passed back as this is essentially just a secHTTP wrapper.

pkg/apiroutes/templates_test.go
```diff
@@ -478,3 +480,38 @@ func TestBatchPatchTemplateRepos(t *testing.T) {
 
 	// This test block cleans up after itself, assuming that the template repo tested was initially enabled. (This test block resets it to 'enabled')
 }
+
+func Test_HTTPRequestWithRetryOnLock(t *testing.T) {
```
jcockbain: not the format of other test names... no underscore

pkg/apiroutes/templates_test.go
```diff
@@ -478,3 +480,38 @@ func TestBatchPatchTemplateRepos(t *testing.T) {
...
+	t.Run("Checks that the 423 is returned if the response StatusCode is always 423", func(t *testing.T) {
```
jcockbain: nit: StatusCode in this message, status code in the next

pkg/apiroutes/templates.go
```diff
@@ -355,3 +357,23 @@ func BatchPatchTemplateRepos(conID string, operations []RepoOperation) ([]SubRes
 
 	return subResponsesFromBatchOperation, nil
 }
+
+// HTTPRequestWithRetryOnLock : Retries a request until either the maxHTTPRetries limit has been hit or the response status code is not equal to 423 (templates locked HTTP code)
+func HTTPRequestWithRetryOnLock(httpClient utils.HTTPClient, originalRequest *http.Request, connection *connections.Connection) (*http.Response, *sechttp.HTTPSecError) {
+	// maxHTTPRequests is the total number of requests, 1 for the intial and the rest are retries
```
jcockbain: typo: initial

pkg/apiroutes/templates_test.go
jcockbain: just "checks 423 is returned"

pkg/apiroutes/templates.go
```diff
@@ -355,3 +357,23 @@ func BatchPatchTemplateRepos(conID string, operations []RepoOperation) ([]SubRes
 
 	return subResponsesFromBatchOperation, nil
 }
+
+// HTTPRequestWithRetryOnLock : Retries a request until either the maxHTTPRetries limit has been hit or the response status code is not equal to 423 (templates locked HTTP code)
+func HTTPRequestWithRetryOnLock(httpClient utils.HTTPClient, originalRequest *http.Request, connection *connections.Connection) (*http.Response, *sechttp.HTTPSecError) {
```
jcockbain: as this is a useful utility function with apiroutes, should it take maxRetries as a param?

pkg/apiroutes/templates.go
```diff
@@ -355,3 +357,23 @@ func BatchPatchTemplateRepos(conID string, operations []RepoOperation) ([]SubRes
...
+		if httpSecError != nil {
+			return nil, httpSecError
+		} else if resp.StatusCode != templatesLockedHTTPStatusCode || i+1 == maxHTTPRequests {
+			response = resp
+			break
+		} else {
+			time.Sleep(time.Second)
+		}
+	}
+	return response, nil
```
jcockbain: will it ever hit this?

pkg/apiroutes/templates.go
james-wallis: It could be enhanced to be this way but at the moment its set at 5 as that is what Toby and I agreed on, if its going to be used as a utility param in the future and we need more or less than 5 then yes it will need to be changed.

pkg/apiroutes/templates.go
jcockbain: can see why it's in though, as you need a response at the end of the function

james-wallis: yes its the only return that returns the response

PR #405, opened 2020-03-11
feat, test: allow insecure keyring mode

rwalle61:
Signed-off-by: Richard Waller <Richard.Waller@ibm.com>

## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?
Implements the functionality requested in https://github.com/eclipse/codewind/issues/1735#issuecomment-593967702:
- [x] No IDE
```
export ALLOW_INSECURE_KEYRING=true
cwctl templates list --conid $CONID
```
- [x] No IDE
```
cwctl --use-insecure-keyring templates list --conid $CONID
```
- [x] IDE (e.g. theia)
```
export ALLOW_INSECURE_KEYRING=true
yarn theia start /home/project --hostname=0.0.0.0
# ide should now leverage this envVar when running cwctl
```

I'm implementing part 1 and 3 now

for https://github.com/eclipse/codewind/issues/1735

yes

### DIFF
--- a/pkg/security/keychain.go
+++ b/pkg/security/keychain.go
@@ -83,7 +83,7 @@ func StoreSecretInKeyring(connectionID, uName, pass string) *SecError {
 	if globals.UseInsecureKeyring {
 		_, statErr := os.Stat(GetPathToInsecureKeyring())
 		if os.IsNotExist(statErr) {
-			os.MkdirAll(insecureKeyringDir, 0777)
+			os.MkdirAll(insecureKeyringDir, 0600)
 			os.OpenFile(GetPathToInsecureKeyring(), os.O_CREATE, 0666)
 		}
 
--- a/pkg/security/keychain.go
+++ b/pkg/security/keychain.go
@@ -84,7 +84,7 @@ func StoreSecretInKeyring(connectionID, uName, pass string) *SecError {
 		_, statErr := os.Stat(GetPathToInsecureKeyring())
 		if os.IsNotExist(statErr) {
 			os.MkdirAll(insecureKeyringDir, 0600)
-			os.OpenFile(GetPathToInsecureKeyring(), os.O_CREATE, 0666)
+			os.OpenFile(GetPathToInsecureKeyring(), os.O_CREATE, 0600)
 		}
 
 		existingSecrets := []KeyringSecret{}
--- a/pkg/security/keychain.go
+++ b/pkg/security/keychain.go
@@ -83,8 +83,14 @@ func StoreSecretInKeyring(connectionID, uName, pass string) *SecError {
 	if globals.UseInsecureKeyring {
 		_, statErr := os.Stat(GetPathToInsecureKeyring())
 		if os.IsNotExist(statErr) {
-			os.MkdirAll(insecureKeyringDir, 0600)
-			os.OpenFile(GetPathToInsecureKeyring(), os.O_CREATE, 0600)
+			mkdirErr := os.MkdirAll(insecureKeyringDir, 0600)
+			if mkdirErr != nil {
+				return &SecError{errOpKeyring, mkdirErr, mkdirErr.Error()}
+			}
+			_, openFileErr := os.OpenFile(GetPathToInsecureKeyring(), os.O_CREATE, 0600)
+			if openFileErr != nil {
+				return &SecError{errOpKeyring, openFileErr, openFileErr.Error()}
+			}
 		}
 
 		existingSecrets := []KeyringSecret{}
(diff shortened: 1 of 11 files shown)


pkg/security/keychain.go
```diff
@@ -58,14 +67,176 @@ func SecKeyUpdate(connectionID string, username string, password string) *SecErr
...
+}
+
+// StoreSecretInKeyring stores the secret in either the system keyring or our insecure keyring.
+func StoreSecretInKeyring(connectionID, uName, pass string) *SecError {
+	service := connectionIDToService(connectionID)
+	if globals.UseInsecureKeyring {
+		_, statErr := os.Stat(GetPathToInsecureKeyring())
+		if os.IsNotExist(statErr) {
+			os.MkdirAll(insecureKeyringDir, 0777)
+			os.OpenFile(GetPathToInsecureKeyring(), os.O_CREATE, 0666)
```
hhellyer: Do you want to give permissions to make this world readable? I think you might be able to stick to user readable only with 0600.

pkg/security/keychain.go
```diff
@@ -58,14 +67,176 @@ func SecKeyUpdate(connectionID string, username string, password string) *SecErr
...
+	// else delete from system keyring
+	err := keyring.Delete(service, uName)
+	if err != nil {
+		return &SecError{errOpKeyring, err, err.Error()}
+	}
+	return nil
+}
+
+func readInsecureKeyring() ([]KeyringSecret, *SecError) {
+	file, readErr := ioutil.ReadFile(GetPathToInsecureKeyring())
```
hhellyer: What does `ioutil.ReadFile()` do if the file doesn't exist? Will that error?

rwalle61: yea `readErr` is `open /Users/richard.waller@ibm.com/.codewind/config/insecureKeychain.json: no such file or directory`. I've just checked this and this automated test also verifies this: https://github.com/eclipse/codewind-installer/pull/405/files#diff-5692282167dc222ced4f1485c91cbad7R87. Whoever calls `readInsecureKeyring` will have to handle the err just like any other

pkg/security/keychain.go
```diff
@@ -58,14 +67,176 @@ func SecKeyUpdate(connectionID string, username string, password string) *SecErr
...
+		_, statErr := os.Stat(GetPathToInsecureKeyring())
+		if os.IsNotExist(statErr) {
+			os.MkdirAll(insecureKeyringDir, 0600)
+			os.OpenFile(GetPathToInsecureKeyring(), os.O_CREATE, 0600)
+		}
 
-	secret, err := keyring.Get(KeyringServiceName+"."+conID, uName)
+		existingSecrets := []KeyringSecret{}
+		file, readErr := ioutil.ReadFile(GetPathToInsecureKeyring())
+		if readErr != nil {
```
hhellyer: If the keyring file doesn't exist the first time you try to store a secret will this function error instead of creating the file?

pkg/security/keychain.go
hhellyer: You make the same call when writing secrets, see: https://github.com/eclipse/codewind-installer/pull/405#discussion_r392221978 - can you write if the keychain file doesn't exist?

pkg/security/keychain.go
rwalle61: it shouldn't error, `len(file)` should be 0 and we will just write the new secret to the file. This test covers it https://github.com/eclipse/codewind-installer/pull/405/files/49e111dcb2302776e1b208b087023ef33f4ac31c#diff-5692282167dc222ced4f1485c91cbad7R94-R98

pkg/security/keychain.go
rwalle61: yep https://github.com/eclipse/codewind-installer/pull/405#discussion_r392240264

rwalle61:
This works for me locally and passes tests locally but the tests are failing in Jenkins. But Jenkins is passing builds even when tests fail right now!

So I am investigating why this passes tests locally but not in Jenkins

rwalle61:
I think the failing tests are just the ones introduced in this PR.

I think some of them (`TestDispatchHTTPRequest`) are failing in Jenkins because they need a system keyring, which Jenkins doesn't have. Mac does have one, so they pass locally for me. So I'm skipping them to see what else is still failing

rwalle61:
The tests using my new insecure keyring are failing too in Jenkins. It can't find the `insecureKeychain.json` we should write out, so maybe we aren't writing successfully in Jenkins. 

So I'm investigating whether we can actually write out any files (e.g. `connections.json`) and getting this to work if possible

PR #410, opened 2020-03-17
Update NOTICE.md

josephca:
Port update from https://github.com/eclipse/codewind/pull/2493

## What type of PR is this ? 

- [ ] Bug fix
- [ ] Enhancement


hhellyer:
@NikCanvin - Do we need this change in the installer NOTICE.md, should we also put an identical PR into the master branch to keep master and 0.10.0 in sync?

PR #412, opened 2020-03-18
Pass docker config remerge

hhellyer:
## What type of PR is this ? 

- [ ] Bug fix
- [X] Enhancement
## What does this PR do ?

This PR adds support for persisting the docker registry credentials between Codewind restarts.
It stores the docker logins added via the 'registrysecrets' command for local connections in the local keychain.
In the docker compose file they are mounted into the docker codewind-pfe container as a text file based secret.
The text file containing the secret is cleared on shutdown or startup failure and is read only to the user to try to provide as much protection as possible. (Unfortunately writing the docker secrets back out to the file system seems unavoidable.) The secret is also base64 encoded, this matches what happens in kubernetes and just protects passwords from accidental casual inspection, it is not intended as a substitute for encryption.
The docker compose file version is updated to 3.3 as that supports secrets and is also the level supported by docker 17.06 which is the minimum version we pre-req:

https://docs.docker.com/compose/compose-file/#secrets (Docker levels and docker compose file versions.)
https://www.eclipse.org/codewind/mdt-vsc-getting-started.html (Docker level required for codewind.)

This PR also includes test cases.

This is part of https://github.com/eclipse/codewind/issues/1306

https://github.com/eclipse/codewind/issues/1306

Note that I still need to raise one further PR to log the local user (outside the PFE container on the host machine) into their registry with the provided docker credentials so that appsody can also pull images. This PR (and it's companion) just solve persisting docker credentials across PFE restarts.

This PR requires the change in https://github.com/eclipse/codewind/pull/2417 for PFE to actually use the secret that is created.
(They can be merged independently but can't be tested without each other as one creates the secret and the other reads it.)

This is a rebase of https://github.com/eclipse/codewind-installer/pull/394 on top of the changes to add an insecure keyring. A change to use the insecure keyring in the codewind builds:
https://github.com/eclipse/codewind/pull/2507 *must* be merged first or this change will break the master build on the codewind repository.

### DIFF
--- a/pkg/security/keychain.go
+++ b/pkg/security/keychain.go
@@ -156,12 +156,16 @@ func GetSecretFromKeyring(connectionID, uName string) (string, *SecError) {
 				return string(secret.Password), nil
 			}
 		}
-		err := errors.New("secret not found in keyring")
+		err := errors.New(textSecretNotFound)
 		return "", &SecError{errOpKeyring, err, err.Error()}
 	}
 	// else get from system keyring
 	secret, err := keyring.Get(service, uName)
 	if err != nil {
+		if err == keyring.ErrNotFound {
+			err := errors.New(textSecretNotFound)
+			return "", &SecError{errOpKeyring, err, err.Error()}
+		}
 		return "", &SecError{errOpKeyring, err, err.Error()}
 	}
 	return secret, nil
@@ -218,6 +222,10 @@ func DeleteSecretFromKeyring(connectionID, uName string) *SecError {
 func readInsecureKeyring() ([]KeyringSecret, *SecError) {
 	file, readErr := ioutil.ReadFile(GetPathToInsecureKeyring())
 	if readErr != nil {
+		if os.IsNotExist(readErr) {
+			err := errors.New(textSecretNotFound)
+			return nil, &SecError{errOpKeyring, err, err.Error()}
+		}
 		return nil, &SecError{errOpKeyring, readErr, readErr.Error()}
 	}
 	secrets := []KeyringSecret{}
--- a/pkg/security/keychain.go
+++ b/pkg/security/keychain.go
@@ -163,8 +163,8 @@ func GetSecretFromKeyring(connectionID, uName string) (string, *SecError) {
 	secret, err := keyring.Get(service, uName)
 	if err != nil {
 		if err == keyring.ErrNotFound {
-			err := errors.New(textSecretNotFound)
-			return "", &SecError{errOpKeyring, err, err.Error()}
+			errNotFound := errors.New(textSecretNotFound)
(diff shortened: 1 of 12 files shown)


pkg/security/keychain.go
```diff
@@ -156,12 +156,16 @@ func GetSecretFromKeyring(connectionID, uName string) (string, *SecError) {
...
-		err := errors.New("secret not found in keyring")
+		err := errors.New(textSecretNotFound)
 		return "", &SecError{errOpKeyring, err, err.Error()}
 	}
 	// else get from system keyring
 	secret, err := keyring.Get(service, uName)
 	if err != nil {
+		if err == keyring.ErrNotFound {
+			err := errors.New(textSecretNotFound)
+			return "", &SecError{errOpKeyring, err, err.Error()}
```
rwalle61: I'm assuming in go this `err` refers to the one defined on the line above, not the one defined 3 lines above

rwalle61:
Thanks for that change. 

The PR build seems to be failing because [a test in filesystem_test.go](https://github.com/eclipse/codewind-installer/blob/master/pkg/docker/filesystem_test.go#L26) calls `WriteToComposeFile`, which [now calls `writeDockerConfigSecretFile`](https://github.com/eclipse/codewind-installer/pull/412/files#diff-5b399f2632e044211e7ee9dcb1f04928R35), which calls [`getDockerCredentials`](https://github.com/eclipse/codewind-installer/pull/412/files#diff-5b399f2632e044211e7ee9dcb1f04928R62), which calls [`security.GetSecretFromKeyring`](https://github.com/eclipse/codewind-installer/pull/412/files#diff-5b399f2632e044211e7ee9dcb1f04928R62), which calls [`keyring.Get`](https://github.com/eclipse/codewind-installer/blob/master/pkg/security/keychain.go#L163), which errors in Jenkins because Jenkins does not have a system keyring. (This test passes but local mac does have a system keyring).

If we set `globals.SetUseInsecureKeyring(true)`[eg](https://github.com/eclipse/codewind-installer/blob/master/pkg/sechttp/httpclient_test.go#L63) then `security.GetSecretFromKeyring` will not call `keyring.Get`, but rather [use our insecure keyring](https://github.com/eclipse/codewind-installer/blob/master/pkg/security/keychain.go#L146). However, it will then [call `readInsecureKeyring`](https://github.com/eclipse/codewind-installer/blob/master/pkg/security/keychain.go#L146) and [receive a `readErr` because the `insecureKeyring.json` does not exist](https://github.com/eclipse/codewind-installer/blob/master/pkg/security/keychain.go#L221). 

Not sure what the solution is. For the past few days I've been trying to get our tests to create the `insecureKeyring.json` in Jenkins, but have not succeeded. Maybe in our Jenkins script we could create that file manually, (maybe that would succeed even though our tests couldn't write it out, or maybe not)

PR #413, opened 2020-03-19
Add project restart command

jcockbain:

## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?

Wraps the PFE rest API in a CWCTL command, with flags for connectionID, projectID and startMode.

At this stage, this is for only local projects. Further PRs will be made, to PFE and this repo, for handling remote project restart in run and debug mode.

Part of https://github.com/eclipse/codewind/issues/1990.

No

### DIFF
--- a/pkg/actions/commands.go
+++ b/pkg/actions/commands.go
@@ -208,7 +208,7 @@ func Commands() {
 					Usage: "Restart a single project, requires 'id' and 'startMode'",
 					Flags: []cli.Flag{
 						cli.StringFlag{Name: "id,i", Usage: "Project ID", Required: true},
-						cli.StringFlag{Name: "startmode, s", Usage: "Start Mode, either run or debug", Required: true},
+						cli.StringFlag{Name: "startmode, s", Usage: "Start Mode of the project; can be run, debug, or debugNoInit", Required: true},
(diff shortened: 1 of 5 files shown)


pkg/actions/commands.go
```diff
@@ -203,6 +203,19 @@ func Commands() {
 						return nil
 					},
 				},
+				{
+					Name:  "restart",
+					Usage: "Restart a single project, requires 'id' and 'startMode'",
+					Flags: []cli.Flag{
+						cli.StringFlag{Name: "id,i", Usage: "Project ID", Required: true},
+						cli.StringFlag{Name: "startmode, s", Usage: "Start Mode, either run or debug", Required: true},
```
markcor11: Cant see where startmode value is being checked.  The usage says it must be either 'run' or 'debug' but the request is dispatched to PFE without validation.  Also seems there is a 3rd allowed option `debugNoInit`.

jcockbain: It gets checked in the request validation of PFE (400 error code if not), I didn't want to add validation in here too in case for example an extra startMode was added and we'd then have to make simultaneous changes. 

Good spot on debugNoInit, that is a 3rd valid mode yep.

PR #414, opened 2020-03-23
Log into docker registries locally when saving credentials.

hhellyer:
## What type of PR is this ? 

- [ ] Bug fix
- [X] Enhancement

## What does this PR do ?
Performs a local docker login when saving docker credentials to support extensions that need access to docker registries.

https://github.com/eclipse/codewind/issues/1306

https://github.com/eclipse/codewind/issues/1306

Yes, we should note that when a user adds a docker registry to the local connection they will be logged into that docker repository on their local machine as well. The documentation may currently assume that local codewind does not save docker registries. The work item for the whole piece of work https://github.com/eclipse/codewind/issues/1306 has been tagged with area/docs so this should be included.

### DIFF
--- a/pkg/actions/registrysecrets.go
+++ b/pkg/actions/registrysecrets.go
@@ -45,22 +45,33 @@ func AddRegistrySecret(c *cli.Context) {
 	username := strings.TrimSpace(c.String("username"))
 	password := strings.TrimSpace(c.String("password"))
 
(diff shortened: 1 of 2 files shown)


hhellyer:
I've amended this to resolve https://github.com/eclipse/codewind/issues/2582 (I raised that thinking this had already been merged.)

PR #419, opened 2020-03-30
Add insecure keyring error code + Restructure start.go error handling (Adds JSON output)

james-wallis:

## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?
* Adds a `sec_insecure_keyring` error to the Keyring errors to make it clearer whether an error has occurred while operating on a Keyring or our insecure keyring file.
    * This means that the IDE's can assume all `sec_keyring` errors should point the user to check their Keyring is installed and working.
* Also adds a `sec_keyring_secret_not_found` error that the IDE's can use to distinguish whether a secret has just not been able to be found rather than indicating a problem reading/writing to the system Keychain.

* Changes the `start.go` command to handle errors using the `HandleDockerError` function rather than using `errors.checkErr` which appears to be an old implementation. This means that if the start command hits an error and the `--json` flag is given we will give back correct JSON as the **final** line of output.
   * Due to the nature of our output this is the best we can do without changing how `cwctl start` behaves.
See here:
```
> cwctl --json start
Debug: false
==> created file /app/root/.codewind/docker-compose.yaml
{"error":"sec_keyring","error_description":"exec: \"dbus-launch\": executable file not found in $PATH"}

> cwctl start
Debug: false
==> created file /app/root/.codewind/docker-compose.yaml
ERRO[0000] exec: "dbus-launch": executable file not found in $PATH
```
### Testing
* Fixed the Keychain tests.
* Fixed any other tests that were broken by my changes to the `start` command

If you want to test this PR with a system with no keychain use the Docker container specified below:
Dockerfile:
```yaml
FROM ubuntu

WORKDIR /app

RUN apt-get -y update 

RUN apt-get -y install docker.io

ADD ./cmd/cli/cwctl-linux /app/cwctl
```

Bash script to be run from root dir in the `codewind-installer` directory on your machine:
```bash
#!/bin/bash

cd cmd/cli
CGO_ENABLED=0 GOOS=linux go build -ldflags="-s -w" -o cwctl-linux 
cd ../..

docker build -t cwctl-insecure .

rm cmd/cli/cwctl-linux

docker run -it -v /var/run/docker.sock:/var/run/docker.sock --rm cwctl-insecure ./cwctl --json start
```


pkg/utils/filesystem.go
```diff
@@ -34,20 +34,22 @@ import (
...
+
 		defer file.Close()
 
 		dir, _ := os.Getwd()
-		fmt.Println("==> created file", path.Join(dir, filePath))
-		return true
+		return path.Join(dir, filePath), nil
 	}
-	return false
+	return "", nil
```
hhellyer, line 52: I'm not sure what this return case means. Is it an error or does it just mean the file already existed and it's ok to re-use it? (Similar to above, can this function now get away with just returning an error or nil. Does it need to return a string?)

PR #433, opened 2020-04-16
2619 - add custom error message for invalid credentials

jcockbain:

## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?

Adds a custom error for if the docker login fails. Examples of this:

with `--json`:
```
cwctl --json --insecure registrysecrets add --conid local --address fakedomain0553.io --username fake-user --password *********
{"error":"DOCKER_CREDENTIAL_ERROR","error_description":"Docker login to fakedomain0553.io as fake-user failed. Address, username or password is incorrect"}
exit status 1
```

without: 
```
cwctl --insecure registrysecrets add --conid local --address fakedomain0553.io --username fake-user --password *********
ERRO[0000] Docker login to fakedomain0553.io as fake-user failed. Address, username or password is incorrect
exit status 1
```

https://github.com/eclipse/codewind/issues/2619

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -16,7 +16,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
-	goErr "errors"
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -357,7 +357,7 @@ func ValidateImageDigest(dockerClient DockerClient, image string) (string, *Dock
 					logr.Tracef("Validation for image digest ..%v succeeded\n", last10)
 				} else {
 					logr.Traceln("Local image digest did not match queried image digest from dockerhub - This could be a result of a bad download")
-					valError := goErr.New(textBadDigest)
+					valError := errors.New(textBadDigest)
 					return image.ID, &DockerError{errOpValidate, valError, valError.Error()}
 				}
 			}
@@ -623,7 +623,7 @@ func LoginToRegistry(address string, username string, password string) *DockerEr
 		return &DockerError{errDockerCredential, err, "Error executing 'docker login'"}
 	}
 
-	// Write to stdin using an asyncrhonous go routine.
+	// Write to stdin using an asynchronous go routine.
 	go func() {
 		defer stdin.Close()
 		io.WriteString(stdin, password)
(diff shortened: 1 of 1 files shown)


pkg/docker/docker.go
```diff
@@ -619,19 +619,22 @@ func LoginToRegistry(address string, username string, password string) *DockerEr
 	// Pipe the password via stdin.
 	cmd := exec.Command("docker", "login", "--username", username, "--password-stdin", address)
 	stdin, err := cmd.StdinPipe()
+	textDockerLoginFailed := fmt.Sprintf("Docker login to %s as %s failed. Address, username or password is incorrect", address, username)
```
james-wallis, line 622: Guess this can't go in the `docker_error.go` file as you want to pass in the username and address?

pkg/docker/docker.go
```diff
@@ -619,19 +619,22 @@ func LoginToRegistry(address string, username string, password string) *DockerEr
...
 	if err != nil {
-		return &DockerError{errDockerCredential, err, "Error executing 'docker login'"}
+		loginErr := errors.New(textDockerLoginFailed)
+		return &DockerError{errDockerCredential, loginErr, textDockerLoginFailed}
```
james-wallis, line 625: Not sure if it would be better to send the original error `err` and then the text here so we can debug this easier?

pkg/docker/docker.go
```diff
@@ -619,19 +619,22 @@ func LoginToRegistry(address string, username string, password string) *DockerEr
...
 	go func() {
 		defer stdin.Close()
 		io.WriteString(stdin, password)
 	}()
 
 	_, err = cmd.CombinedOutput()
 	if err != nil {
-		return &DockerError{errDockerCredential, err, "Error executing 'docker login'"}
+		loginErr := errors.New(textDockerLoginFailed)
+		return &DockerError{errDockerCredential, loginErr, textDockerLoginFailed}
```
james-wallis, line 637: same as above

pkg/docker/docker.go
jcockbain: That's why it's in the function yes

pkg/docker/docker.go
jcockbain: That's the cause of the original issue - err returned from `os/exec` is `exit status 1` for the example inputs. As this is the error message displayed when the `--json` flag is set, that's what is seen in the IDEs.

I'm not sure of a way to include both our custom error, and the original `os/exec` error in the outputs.

PR #434, opened 2020-04-17
2647.1 download template from secure repo

rwalle61:
## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?
`cwctl` can now download templates from GHE repos via `cwctl project create --url <url> --path <path> --username <username> --password <password>`.

Used the `go-github` client to use GitHub's API:
- https://github.com/google/go-github/blob/259aa56b9e3f43b924520fb5b5bc0784a901f90d/github/repos_releases.go
- https://pkg.go.dev/github.com/google/go-github/github?tab=doc#RepositoriesService
- https://developer.github.com/v3/repos/releases/#list-assets-for-a-release

first step of https://github.com/eclipse/codewind/issues/2647

yes, included here

To run the tests that need to pass GitHub auth, you will have to manually input your credentials, because I have not found a way to safely include credentials

### DIFF
--- a/cmd/cli/main_test.go
+++ b/cmd/cli/main_test.go
@@ -19,12 +19,14 @@ import (
 	"testing"
 
 	"github.com/eclipse/codewind-installer/pkg/security"
+	"github.com/eclipse/codewind-installer/pkg/test"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 const cwctlName = "cwctl_test"
 const cwctl = "./" + cwctlName
+const testDir = "./testDir"
 
 func TestCwctl(t *testing.T) {
 	if testing.Short() {
@@ -36,6 +38,7 @@ func TestCwctl(t *testing.T) {
 
 	testBasicUsage(t)
 	testUseInsecureKeyring(t)
+	testCreateProjectFromTemplate(t)
 
 	os.Remove(cwctlName)
 }
@@ -104,5 +107,52 @@ func testUseInsecureKeyring(t *testing.T) {
 		require.Equal(t, "seCretphrase", string(secret.Password))
 
 		os.Remove(security.GetPathToInsecureKeyring())
+	})
+}
+
+func testCreateProjectFromTemplate(t *testing.T) {
+	t.Run("cwctl project create --url <insecureTemplateRepo> --path <testDir>", func(t *testing.T) {
+		os.RemoveAll(testDir)
+		defer os.RemoveAll(testDir)
+
+		cmd := exec.Command(cwctl, "project", "create",
+			"--url="+test.PublicGHRepoURL,
+			"--path="+testDir,
+		)
+		out, err := cmd.Output()
+		assert.Nil(t, err)
+		assert.Equal(t, "{\"status\":\"success\",\"projectPath\":\"./testDir\",\"result\":{\"language\":\"javascript\",\"projectType\":\"nodejs\"}}\n", string(out))
+	})
+	t.Run("cwctl project create --url <secureTemplateRepo> --path <testDir> --username <test.GHEUsername> --password <test.GHEPassword>", func(t *testing.T) {
+		if !test.UsingOwnGHECredentials {
+			t.Skip("skipping this test because you haven't set GitHub credentials needed for this test")
+		}
+
+		os.RemoveAll(testDir)
+		defer os.RemoveAll(testDir)
+
+		cmd := exec.Command(cwctl, "project", "create",
+			"--url="+test.GHERepoURL,
+			"--path="+testDir,
+			"--username="+test.GHEUsername,
+			"--password="+test.GHEPassword,
+		)
+		out, err := cmd.Output()
+		assert.Nil(t, err)
+		assert.Equal(t, "{\"status\":\"success\",\"projectPath\":\"./testDir\",\"result\":{\"language\":\"unknown\",\"projectType\":\"docker\"}}\n", string(out))
+	})
+	t.Run("cwctl project create --url <secureTemplateRepo> --path <testDir> --username <goodUsername> --password <badPassword>", func(t *testing.T) {
+		os.RemoveAll(testDir)
+		defer os.RemoveAll(testDir)
+
+		cmd := exec.Command(cwctl, "project", "create",
+			"--url="+test.GHERepoURL,
+			"--path="+testDir,
+			"--username="+test.GHEUsername,
+			"--password=badpassword",
+		)
+		out, err := cmd.Output()
+		assert.NotNil(t, err)
+		assert.Equal(t, "", string(out))
 	})
 }
--- a/cmd/cli/main_test.go
+++ b/cmd/cli/main_test.go
@@ -19,6 +19,7 @@ import (
 	"testing"
 
 	"github.com/eclipse/codewind-installer/pkg/security"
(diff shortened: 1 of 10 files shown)


pkg/utils/download.go
```diff
@@ -12,24 +12,27 @@
...
 // DownloadFromURLThenExtract downloads files from a URL
 // to a destination, extracting them if necessary
-func DownloadFromURLThenExtract(URL string, destination string) error {
+func DownloadFromURLThenExtract(URL, destination string, gitCredentials GitCredentials) error {
 	if _, err := url.ParseRequestURI(URL); err != nil {
 		return err
 	}
 
 	if IsTarGzURL(URL) {
 		return DownloadFromTarGzURL(URL, destination)
```
hhellyer, line 33: Do we need to supply credentials when downloading a tar.gz? (I can't remember which project type uses this instead of zip, it may be Appsody.)

cmd/cli/main_test.go
```diff
@@ -86,22 +90,69 @@ func testUseInsecureKeyring(t *testing.T) {
...
+
+		cmd := exec.Command(cwctl, "project", "create",
+			"--url="+test.InsecureGitRepo,
+			"--path="+testDir,
+		)
+		out, err := cmd.Output()
+		assert.Nil(t, err)
+		assert.Equal(t, "{\"status\":\"success\",\"projectPath\":\"./testDir\",\"result\":{\"language\":\"javascript\",\"projectType\":\"nodejs\"}}\n", string(out))
+	})
+	t.Run("cwctl project create --url <secureTemplateRepo> --path <testDir> --username <test.GoodGitHubUsername> --password <test.GoodGitHubPassword>", func(t *testing.T) {
```
jcockbain, line 126: Trying to run this now with my credentials and getting a 404 error. Debugging but may help to have some more information somewhere about configuring the tests. 

Perhaps we should add some extra to the README / contribution guides about running the tests - this could go in there alongside details about running with `--short` etc..

pkg/utils/download.go
rwalle61: I didn't realise - should I make that change in this or the next PR?

cmd/cli/main_test.go
rwalle61: Are you using your GHE credentials, not GH credentials? Because the sample secure repo is GHE. GH creds would work for a private GH repo

jcockbain: I changed the secure repo to a private GH repo, and used my GH credentials

pkg/project/create.go
```diff
@@ -71,7 +70,7 @@ func DownloadTemplate(destination string, url string) (*Result, *ProjectError) {
 		projectName = "PROJ_NAME_PLACEHOLDER"
 	}
 
-	err := utils.DownloadFromURLThenExtract(url, destination)
+	err := utils.DownloadFromURLThenExtract(url, destination, gitCredentials)
```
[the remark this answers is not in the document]
rwalle61: think it's type `string` in both places, but named URL in the function param

cmd/cli/main_test.go
rwalle61: ah yes, thanks, turns out I need to make a change to get it to download private GH repos

pkg/test/globals.go
```diff
@@ -0,0 +1,27 @@
...
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+
+package test
+
+// InsecureGitRepo is a URL to a GitHub repo not requiring auth to access
+const InsecureGitRepo = "https://github.com/microclimate-dev2ops/nodeExpressTemplate"
+
+// SecureGitRepo is a URL to a GitHub repo requiring auth to access (e.g. a GitHub Enterprise repo or a private GitHub repo)
+const SecureGitRepo = "https://github.ibm.com/Richard-Waller/git-basics"
```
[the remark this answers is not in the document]
rwalle61: I'll use a different GHE repo

pkg/utils/download.go
```diff
@@ -48,33 +74,119 @@ func DownloadFromTarGzURL(URL string, destination string) error {
 	return err
 }
 
+func getURLToDownloadReleaseAsset(URL string, gitCredentials GitCredentials) (string, error) {
+	URLSlice := strings.Split(URL, "/")
```
hhellyer, line 78: Would it be safer to parse this as a URL and use `URL.Hostname` see: https://golang.org/pkg/net/url/#URL.Hostname
We're already using the `net/url` package elsewhere in download.go.
This probably won't quite give you what you want if there's a port in the hostname for example.

pkg/utils/download.go
```diff
@@ -12,34 +12,60 @@
...
+		Password string `json:"password"`
+	}
 )
 
 // DownloadFromURLThenExtract downloads files from a URL
 // to a destination, extracting them if necessary
-func DownloadFromURLThenExtract(URL string, destination string) error {
-	if _, err := url.ParseRequestURI(URL); err != nil {
+func DownloadFromURLThenExtract(URL, destination string, gitCredentials GitCredentials) error {
+	_url, err := url.ParseRequestURI(URL)
```
hhellyer, line 39: I'm assuming `_url` was picked as a name to avoid a clash with `url`.

pkg/utils/download.go
```diff
@@ -48,33 +74,119 @@ func DownloadFromTarGzURL(URL string, destination string) error {
 	return err
 }
 
+func getURLToDownloadReleaseAsset(URL string, gitCredentials GitCredentials) (string, error) {
+	URLSlice := strings.Split(URL, "/")
+
+	domain := URLSlice[2]
```
hhellyer, line 80: Should there be a check on the length of URLSlice? We use up to index 7 without verifying it has 7 elements.

pkg/utils/download.go
```diff
@@ -48,33 +74,119 @@ func DownloadFromTarGzURL(URL string, destination string) error {
...
 // DownloadFromRepoURL downloads a repo from a URL to a destination
-func DownloadFromRepoURL(repoURL string, destination string) error {
-	// expecting string in format 'https://github.com/<owner>/<repo>'
-	if strings.HasPrefix(repoURL, "https://") {
-		repoURL = strings.TrimPrefix(repoURL, "https://")
+func DownloadFromRepoURL(repoURL, destination string, gitCredentials GitCredentials) error {
+	// expecting string in format 'https://<github.com>/<owner>/<repo>'
+	URLSlice := strings.Split(repoURL, "/")
+
+	domain := URLSlice[2]
```
hhellyer, line 138: We should probably check the length of URLSlice here too.

pkg/utils/download.go
rwalle61: yep, unfortunately!

pkg/utils/download.go
rwalle61: okay. this is just a private helper function, and we do parse this as a URL above, so I think we're quite safe

But I agree this function would be more reusable if we handled a URL type here, so I can make this change

pkg/utils/download.go
rwalle61: Thanks, I'll add a test case in

pkg/utils/download.go
rwalle61: Thanks, I'll add a test case in

PR #436, opened 2020-04-17
Local `mustgather` cli solution for log gathering for PD.

mattcolegate:
## What type of PR is this ? 

- [ ] Bug fix
- [X] Enhancement

## What does this PR do ?
This PR introduces a new `mustgather` command to the CLI which will collect codewind & project container logs, cwctl verion, docker-compose file, and IDE logs from a Local Codewind and zip them up together to send to the Codewind team for easier problem diagnosis.

Local solution for https://github.com/eclipse/codewind/issues/1579

https://github.com/eclipse/codewind/issues/1579

Yes

This has been successfully tested on both Windows and Mac.

### DIFF
--- a/pkg/actions/mustgather.go
+++ b/pkg/actions/mustgather.go
@@ -1,0 +1,177 @@
+/*******************************************************************************
+ * Copyright (c) 2020 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+
+package actions
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
+	"github.com/eclipse/codewind-installer/pkg/appconstants"
+	"github.com/eclipse/codewind-installer/pkg/docker"
+	"github.com/eclipse/codewind-installer/pkg/errors"
+	"github.com/urfave/cli"
+)
+
+var codewindHome = filepath.Join(os.Getenv("HOMEPATH"), ".codewind")
+var mustGatherDirName = filepath.Join(codewindHome, "mustgather", time.Now().Format("20060102150405"))
+
+func logMG(input ...string) {
+	fmt.Println(input[0])
+}
+
+//MustGatherCommand to gather logs and project files to aid diagnosis of Codewind errors
+func MustGatherCommand(c *cli.Context) {
+	dirErr := os.MkdirAll(filepath.Join(mustGatherDirName, "projects"), 0755)
+	if dirErr != nil {
+		errors.CheckErr(dirErr, 205, "")
+	}
+	// Collect Codewind container inspection & logs
+	for _, cwContainerName := range docker.ContainerNames {
+		logMG("Collecting information from container " + cwContainerName)
+		containerID := getContainerID(cwContainerName)
+		writeContainerInspectToFile(containerID, cwContainerName)
+		writeContainerLogToFile(containerID, cwContainerName)
+	}
+
+	// Collect project container inspection & logs
+	dockerClient, dockerErr := docker.NewDockerClient()
+	if dockerErr != nil {
+		HandleDockerError(dockerErr)
+		os.Exit(1)
+	}
+	allContainers, cListErr := docker.GetContainerListWithOptions(dockerClient, types.ContainerListOptions{All: true})
+	if cListErr != nil {
+		HandleDockerError(cListErr)
+		os.Exit(1)
+	}
+	for _, cwContainer := range docker.GetContainersToRemove(allContainers) {
+		logMG("Collecting information from container " + cwContainer.Names[0])
+		writeContainerInspectToFile(cwContainer.ID, filepath.Join("projects", cwContainer.Names[0]))
+		writeContainerLogToFile(cwContainer.ID, filepath.Join("projects", cwContainer.Names[0]))
+	}
+
+	// Collect docker-compose file
+	copyFileHere(filepath.Join(codewindHome, "docker-compose.yaml"), "docker-compose.yaml")
+
+	// Collect codewind version
+	logMG("Collecting CWCTL version")
+	d1 := []byte(appconstants.VersionNum)
+	versionErr := ioutil.WriteFile(filepath.Join(mustGatherDirName, "cwctl.version"), d1, 0644)
+	if versionErr != nil {
+		errors.CheckErr(versionErr, 201, "")
+	}
+
+	// Attempt to gather VSCode logs
+	//logMG("Collecting VSCode logs")
+	vsCodeLogsDir := ""
+	switch os.Getenv("OSTYPE") {
+	case "darwin":
+		vsCodeLogsDir = filepath.Join(os.Getenv("HOME"), "Library", "Application Support", "Code", "logs")
+	case "linux-gnu":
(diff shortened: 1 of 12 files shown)


mattcolegate:
@johnmcollier This is the Local only solution so that the team can use it internally for early feedback.

pkg/actions/mustgather.go
```diff
@@ -0,0 +1,364 @@
...
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
+	"github.com/eclipse/codewind-installer/pkg/appconstants"
+	"github.com/eclipse/codewind-installer/pkg/docker"
+	"github.com/eclipse/codewind-installer/pkg/errors"
+	"github.com/urfave/cli"
+)
+
+var codewindHome = filepath.Join(homeDir, ".codewind")
+var nowTime = time.Now().Format("20060102150405")
```
hhellyer, line 36: I was going to comment on this constant. Then I learnt about go's "magical reference date"...

mattcolegate: Crazy isn't it!

pkg/actions/mustgather.go
```diff
@@ -0,0 +1,364 @@
...
+		}
+	}
+
+	// Collect docker-compose file
+	logMG("Collecting docker-compose.yaml")
+	copyFileHere(filepath.Join(codewindHome, "docker-compose.yaml"), "docker-compose.yaml")
+
+	// Collect codewind version
+	logMG("Collecting CWCTL version")
+	d1 := []byte(appconstants.VersionNum)
```
hhellyer, line 96: `d1` isn't a great name. Could it be versionNum? (Also there's no d2!)

pkg/actions/mustgather.go
```diff
@@ -0,0 +1,364 @@
...
+			writer.Close()
+		default:
+			logMG("Unable to untar type : " + string(header.Typeflag) + " in file " + filename)
+		}
+	}
+	return nil
+}
+
+//copyFileHere - copies the contents of the source file to a target file in the mustgather directory
+func copyFileHere(sourceFilePath, targetFile string) error {
```
hhellyer, line 342: `copyFileHere` - this takes a location so `Here` is redundant. Also could this move to pkg/utils/filesystem.go?

pkg/actions/mustgather.go
```diff
@@ -0,0 +1,364 @@
...
+	copyCodewindWorkspace(pfeContainerID)
+
+	if c.Bool("projects") {
+		// Collect project container inspection & logs
+		dockerClient, dockerErr := docker.NewDockerClient()
+		if dockerErr != nil {
+			HandleDockerError(dockerErr)
+			os.Exit(1)
+		}
+		allContainers, cListErr := docker.GetContainerListWithOptions(dockerClient, types.ContainerListOptions{All: true})
```
hhellyer, line 78: I'm not sure why `GetContainerListWithOptions` is used here. Is it to pick up stopped containers as well? (If so a comment would be useful just to say why it's different to the logic in StopAllCommand.)

pkg/actions/mustgather.go
```diff
@@ -0,0 +1,364 @@
...
+		if cListErr != nil {
+			HandleDockerError(cListErr)
+			os.Exit(1)
+		}
+		for _, cwContainer := range docker.GetContainersToRemove(allContainers) {
```
hhellyer, line 83: I'd be very tempted to rename `docker.GetContainersToRemove` now to something like `docker.GetCodewindProjectContainers` since we're using it not just for the list of containers to stop.

pkg/actions/mustgather.go
```diff
@@ -0,0 +1,364 @@
...
+	// Attempt to gather VSCode logs
+	logMG("Collecting VSCode logs")
+	vsCodeDir := ""
+	switch runtime.GOOS {
+	case "darwin":
+		vsCodeDir = filepath.Join(homeDir, "Library", "Application Support", "Code")
+	case "linux":
+		vsCodeDir = filepath.Join(homeDir, ".config", "Code")
+	case "windows":
+		vsCodeDir = filepath.Join(homeDir, "AppData", "Roaming", "Code")
```
hhellyer, line 140: Does this need a default case for unsupported OSs?

pkg/actions/mustgather.go
```diff
@@ -0,0 +1,364 @@
...
+var isLoud = true
+
+func logMG(input ...string) {
+	if isLoud {
+		fmt.Println(input[0])
+	}
+}
+
+//MustGatherCommand to gather logs and project files to aid diagnosis of Codewind errors
+func MustGatherCommand(c *cli.Context) {
```
hhellyer, line 48: This function is a bit on the big side. Could it be split up into parts for each stage? For example the Zipping at the end could go into pkg/utils/filesystem.go which already has an `UnZip()` function to unzip a file to a destination.

pkg/actions/mustgather.go
```diff
@@ -0,0 +1,364 @@
...
+		os.Exit(1)
+	}
+	tarFileStream, fileErr := docker.GetFilesFromContainer(dockerClient, containerID, "/codewind-workspace")
+	if fileErr != nil {
+		HandleDockerError(fileErr)
+		os.Exit(1)
+	}
+	defer tarFileStream.Close()
+	// Extracting tarred files
+	tarBallReader := tar.NewReader(tarFileStream)
```
hhellyer, line 297: Could the tar reading part of this be merged with the code we have in filesystem.go UnTar
https://github.com/eclipse/codewind-installer/blob/18d4b2bd1087017b0957ec07932f38e66db578cc/pkg/utils/filesystem.go#L151
then we only have to have the code to handle properly reading a tar file in one place.

pkg/actions/mustgather.go
mattcolegate: Changed to `versionByteArray`

pkg/actions/mustgather.go
mattcolegate: Agreed - moved to pkg/utils/filesystem.go/CopyFile

pkg/actions/mustgather.go
mattcolegate: comment added

pkg/actions/mustgather.go
mattcolegate: Agreed

pkg/actions/mustgather.go
mattcolegate: No - default case already handled at line 135

pkg/actions/mustgather.go
mattcolegate: Sure - I was trying to keep the main function having content as opposed to just being a list of function calls, but I can see how that might make it clearer.

pkg/actions/mustgather.go
mattcolegate: Agreed

mattcolegate:
@hhellyer Requested changes complete

pkg/actions/mustgather.go
hhellyer: True, although if the default case put out the log message at line 198 and then returned you could get rid of `if len(vsCodeDir) > 0 {` and outdent everything by one. (This isn't a change request though, it is fine as it is.)

pkg/actions/mustgather.go
```diff
@@ -0,0 +1,315 @@
...
+)
+
+var codewindHome = filepath.Join(homeDir, ".codewind")
+var nowTime = time.Now().Format("20060102150405")
+var mustGatherMasterDirName = filepath.Join(codewindHome, "mustgather")
+var mustGatherDirName = filepath.Join(mustGatherMasterDirName, nowTime)
+
+var isLoud = true
+
+func logMG(input ...string) {
```
hhellyer, line 41: This function is a bit odd, it takes any number of strings as an argument *but* only prints out the first one. All the callers only ever call it with one argument. Could this be simplified down to:
```
func logMG(input string) {
	if isLoud {
		fmt.Println(input)
	}
}
```
?

mattcolegate: Agreed

PR #443, opened 2020-04-24
feat: 2647.3 `cwctl template repos add` secure repos

rwalle61:
## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?
Extends `cwctl templates repos add` so that it can pass GitHub credentials to PFE, which PFE can handle when https://github.com/eclipse/codewind/pull/2747 is merged

The credentials are sent to PFE in the request body. I haven't encrypted the credentials because I assume we'll be using HTTPS when we care about security. Happy to improve security though if there are any suggestions

See commit messages for more details

Part of https://github.com/eclipse/codewind/issues/2647

included in this PR (see README)

Codecov is failing only because the tests that cover my new code are skipped in Jenkins because they need Git credentials

### DIFF
--- a/pkg/actions/templates.go
+++ b/pkg/actions/templates.go
@@ -68,8 +68,15 @@ func AddTemplateRepo(c *cli.Context) {
 	url := c.String("url")
 	desc := c.String("description")
 	name := c.String("name")
+	username := c.String("username")
+	password := c.String("password")
+
+	gitCredentials := utils.GitCredentials{
+		Username: username,
+		Password: password,
+	}
 	conID := strings.TrimSpace(strings.ToLower(c.String("conid")))
-	repos, err := apiroutes.AddTemplateRepo(conID, url, desc, name)
+	repos, err := apiroutes.AddTemplateRepo(conID, url, desc, name, gitCredentials)
 	if err != nil {
 		templateErr := &TemplateError{errOpAddRepo, err, err.Error()}
 		HandleTemplateError(templateErr)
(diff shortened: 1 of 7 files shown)


pkg/actions/templates.go
```diff
@@ -68,8 +68,15 @@ func AddTemplateRepo(c *cli.Context) {
 	url := c.String("url")
 	desc := c.String("description")
 	name := c.String("name")
+	username := c.String("username")
+	password := c.String("password")
+
+	gitCredentials := utils.GitCredentials{
+		Username: username,
+		Password: password,
```
hhellyer, line 76: Do we need to check if the user has provided both rather than just one or the other? Or do we rely on portal to do that? We do check the URL in `AddTemplateRepo()`. We don't check description but that's optional and doesn't depend on another parameter.

rwalle61: I can add that check in a follow up PR, when we allow the user to authenticate via an access token

Curretnly:
`cwctl templates repos add` just passes whatever username and password it gets to PFE, without checking first what it's received. PFE master does not currently verify that it received both username and password. But my PFE Access Token PR does check this (https://github.com/eclipse/codewind/pull/2762)

hhellyer: It makes sense to do this in the follow up when we have another authentication field to pass that you wouldn't use with the other one. I'll approve and merge this PR.

rwalle61: thanks, I've implemented your suggestion in https://github.com/eclipse/codewind-installer/pull/453

PR #446, opened 2020-04-28
2109 - handle performance container not running when finding PFE host and port

jcockbain:

## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?

Modifies the `CheckContainerStatus` function to include an array of prefixes as a param, and then check if a container with each of these prefixes exists. 

This allows us to check for only `PFE` when determining PFE hostname and port (the cause of the associated issue), but to also check for both `PFE` and `performance` in the `start` and `status` commands. 

Adds extra unit testing for these functions with `performance` running and not running.

https://github.com/eclipse/codewind/issues/2109

### DIFF
--- a/pkg/actions/diagnostics.go
+++ b/pkg/actions/diagnostics.go
@@ -210,7 +210,7 @@ func dgSharedCommand(c *cli.Context) {
 
 // Collect Codewind container inspection & logs
 func collectCodewindContainers() {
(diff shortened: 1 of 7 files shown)


jcockbain:
The decrease in coverage seems to be due to the mock docker clients being included in the codecov stats (and I add an extra mock in the PR). 

Working now to modify the codecov.yml to ignore these functions.

PR #478, opened 2020-05-20
Diagnostics: Make Eclipse/IntelliJ log collection status more clear

mattcolegate:
Signed-off-by: Matthew Colegate <colegate@uk.ibm.com>

## What type of PR is this ? 

- [X] Bug fix
- [ ] Enhancement

## What does this PR do ?
This PR addresses the `cwctl diagnostics` issues described in https://github.com/eclipse/codewind/issues/2857 by:
* Amending the README to remove confusing mentions of defaults for Eclipse and IntelliJ log dirs
* Alters the README for IntelliJ log dirs so that end-users understand they only need to specify it if they are not using the default location
* Removes any mention of collecting Eclipse log dirs if the Eclipse log directory parameter is not specified
* More explicit logging when we are using default IntelliJ log directories .

https://github.com/eclipse/codewind/issues/2857
https://github.com/eclipse/codewind/issues/2857
No

### DIFF
--- a/pkg/actions/diagnostics.go
+++ b/pkg/actions/diagnostics.go
@@ -343,13 +343,31 @@ func gatherCodewindEclipseLogs(codewindEclipseWSDir string) {
 			}
 			logDG("done\n")
 		} else {
-			warnDG("Unable to collect Eclipse logs", "workspace metadata directory not found")
+			warnDG("Unable to collect Eclipse logs", "workspace metadata directory not found in "+codewindEclipseWSDir)
 		}
 	}
 }
 
+func gatherEntireLogDirectory(logDirPath, collectionDirPath, parentLogDirPath string) {
+	err := filepath.Walk(logDirPath, func(path string, info os.FileInfo, err error) error {
+		localPath := filepath.Join(collectionDirPath, strings.Replace(path, parentLogDirPath, "", 1))
+		if info.IsDir() {
+			logDirErr := os.MkdirAll(localPath, 0755)
+			if logDirErr != nil {
+				errors.CheckErr(logDirErr, 205, "")
+			}
+		}
+		if info.Mode().IsRegular() {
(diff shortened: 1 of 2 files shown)


pkg/actions/diagnostics.go
```diff
@@ -339,13 +339,11 @@ func gatherCodewindEclipseLogs(codewindEclipseWSDir string) {
...
 			warnDG("Unable to collect Eclipse logs", "workspace metadata directory not found")
 		}
-	} else {
-		warnDG("Unable to collect Eclipse logs", "workspace not specified")
 	}
 }
 
 func gatherCodewindVSCodeLogs() {
-	logDG("Collecting VSCode logs ... ")
+	logDG("Attempting to collecting VSCode logs from default location ... ")
```
hhellyer, line 346: If you moved this after the switch block you could say what the default location was.

pkg/actions/diagnostics.go
```diff
@@ -409,10 +407,9 @@ func findIntellijDirectory(inDir string) string {
 }
 
 func gatherCodewindIntellijLogs(codewindIntellijLogDir string) {
-	logDG("Collecting Intellij logs ... ")
 	intellijLogsDir := codewindIntellijLogDir
 	if intellijLogsDir == "" {
-		// attempt to use default path
+		logDG("Attempting to collect Intellij logs from default location ... ")
```
hhellyer, line 412: Again if this was after the switch then we could tell the user where we were looking. You could also put it inside the `if len(` block so it was logged even if they specified a directory.

pkg/actions/diagnostics.go
mattcolegate: OK, done

pkg/actions/diagnostics.go
mattcolegate: OK, think I've got those in the right place now!

PR #494, opened 2020-06-10
Fix sync behaviour for imported files

stalleyj:
## What type of PR is this ? 

- [X] Bug fix
- [ ] Enhancement

## What does this PR do ?
This PR fixes our syncing behaviour to identify files that have been imported into the project via the Eclipse IDE.  When a user does this, the modified time of the file does not get updated so our current sync checks for last mod time > last sync time don't pick up the new files.  

The new behaviour looks to see if each file contained in the project (on the users disk) is in the list of files for the project (that have been uploaded to codewind). If that file doesn't exist, it will be uploaded. 

https://github.com/eclipse/codewind/issues/2860

No

No


pkg/project/filelist.go
```diff
@@ -0,0 +1,47 @@
...
+	if readErr != nil {
+		return nil, &ProjectError{errOpRequest, readErr, readErr.Error()}
+	}
+
+	var list FileList
+	jsonErr := json.Unmarshal(byteArray, &list)
+	if jsonErr != nil {
+		return nil, &ProjectError{errOpRequest, jsonErr, jsonErr.Error()}
+	}
+	//	fmt.Println(list)
```
hhellyer, line 45: Does this need tidying? (Commented out `Println`.)

stalleyj: Done

pkg/project/sync.go
```diff
@@ -457,3 +429,79 @@ func handleMissingProjectDir(httpClient utils.HTTPClient, connection *connection
...
+	zWriter.Close()
+	encoded := base64.StdEncoding.EncodeToString(buffer.Bytes())
+	fileUploadBody.Message = encoded
+
+	buf := new(bytes.Buffer)
+	json.NewEncoder(buf).Encode(fileUploadBody)
+
+	projectUploadURL := conURL + "/api/v1/projects/" + projectID + "/upload"
+	// TODO - How do we handle partial success?
+	request, err := http.NewRequest("PUT", projectUploadURL, bytes.NewReader(buf.Bytes()))
```
hhellyer, line 494: I don't think it's likely to happen and this isn't new code but we don't check `err` here.

mattcolegate:
Tested on Windows using recreation instructions from https://github.com/eclipse/codewind/issues/2860 - imported hello/Hello.java and project rebuilt and I was able to run the new application from the /hello endpoint

pkg/project/sync.go
stalleyj: I tried not to change anything if I didn't need to, can be addressed later

PR #495, opened 2020-06-10
Add project name to project link list output

james-wallis:

## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?
Adds the project name into the response for `cwctl project link list --id`.
JSON output:
```
❯ cwctl --json project link list --id 7961a260-a4e1-11ea-8609-af62c8d34aa0
[{"projectID":"e07d2ef0-a0fa-11ea-8c83-95e049ef26c8","projectName":"webby","envName":"DA","projectURL":"cw-webby-e07d2ef0-a0fa-11ea-8c83-95e049ef26c8:9080"}]
```

Human readable output:
```
❯ cwctl project link list --id 7961a260-a4e1-11ea-8609-af62c8d34aa0
PROJECT NAME 	ENVIRONMENT VARIABLE
webby		DA
```

### DIFF
--- a/pkg/actions/project.go
+++ b/pkg/actions/project.go
@@ -291,9 +291,9 @@ func ProjectLinkList(c *cli.Context) {
 		} else {
 			w := new(tabwriter.Writer)
 			w.Init(os.Stdout, 0, 8, 2, '\t', 0)
-			fmt.Fprintln(w, "PROJECT NAME \tENVIRONMENT VARIABLE")
+			fmt.Fprintln(w, "TARGET PROJECT \tENVIRONMENT VARIABLE \t TARGET URL")
 			for _, project := range links {
-				fmt.Fprintln(w, project.ProjectName+"\t"+project.EnvName)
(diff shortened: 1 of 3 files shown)


pkg/actions/project.go
```diff
@@ -291,9 +291,9 @@ func ProjectLinkList(c *cli.Context) {
 		} else {
 			w := new(tabwriter.Writer)
 			w.Init(os.Stdout, 0, 8, 2, '\t', 0)
-			fmt.Fprintln(w, "ENVIRONMENT VARIABLE \tPROJECT ID")
+			fmt.Fprintln(w, "PROJECT NAME \tENVIRONMENT VARIABLE")
```
hhellyer, line 294: Could we add the url here? Maybe as `Target URL` otherwise the only way to get that for debugging is to use the JSON mode.
(Also Target Project might be a better name for the first column to make it clear it's not the current project.)

PR #500, opened 2020-06-24
fix: fix project file mode on windows

rwalle61:
## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?
_fixes https://github.com/eclipse/codewind/issues/2713 by ensuring that project files on Windows are uploaded with the required permissions (775). Previously project files on Windows were uploaded with the wrong permissions, and we cannot detect the correct permissions accurately with Go on Windows https://github.com/eclipse/codewind/issues/2713#issuecomment-625263695

I tested that this works by verifying that setting the file mode in cwctl to be 509 indeed results in the file mode in PFE `codewind-workspace` to be 775 (octal representation of 509).

no

### DIFF
--- a/pkg/project/sync.go
+++ b/pkg/project/sync.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/project/sync.go
```diff
@@ -471,9 +472,13 @@ func syncFile(client utils.HTTPClient, projectID string, projectPath string, pat
 		return uploadResponse
 	}
 
+	mode := uint(fileStat.Mode().Perm())
+	if runtime.GOOS == "windows" {
+    mode = 775
```
hhellyer, line 477: The indentation is off here.

pkg/project/sync.go
```diff
@@ -471,9 +472,13 @@ func syncFile(client utils.HTTPClient, projectID string, projectPath string, pat
 		return uploadResponse
 	}
 
+	mode := uint(fileStat.Mode().Perm())
+	if runtime.GOOS == "windows" {
+		mode = 509
```
hhellyer, line 477: It would be clearer to use octal notation here, that's normal for file permissions. Can you change `509` to `0775` or `0o775`?

